Issue #9, opened 2017-08-09
allow specifying raw values.yaml


calind:
The current approach works well for simple `values.yaml` overrides, but it gets pretty ugly when you need to pass complex yaml. By allowing to specify a custom `values.yaml` as string these complex cases can be templated.

Also I would suggest renaming the curent `value` option with `set` to be more inline with helm cli. `set` would get priority over `values`.

Here is an example for grafana deployment with ingress using the proposed changes:

```tf
data "template_file" "grafana_values" {
  template = <<EOF
server:
  ingress:
    enabled: true
    annotations:
      kubernetes.io/ingress.class: 'nginx'
      kubernetes.io/tls-acme: 'true'
    hosts:
      - grafana.example.com
    tls:
      - secretName: grafana-server-tls
        hosts:
          - grafana.example.com
EOF
}

resource "helm_chart" "grafana" {
  name        = "grafana"
  chart       = "stable/grafana"
  values      = "${data.template_file.grafana_values.rendered}"
  set {
    name    = "server.image"
    content = "grafana/grafana:4.4.3"
  }
}
```

I wil try comming up with a PR.

mcuadros:
For me this goes against Terraform philosophy, but is true that the name/content pair, is a bit tedious, at the very beginning of the development I tried to implement this in a interface{} allowing write HCL more similar to the structure of the values. 

Let me thing about this.

mcuadros:
Implemented at https://github.com/mcuadros/terraform-provider-helm/commit/5d5bad85f183ea7970bb633a0a92454214b66258

PR #15, opened 2017-08-23
helm_chart: allow specifying values as raw yaml

calind:
Adds `raw_values` attribute for helm_chart resource. This allows using raw yaml for values, as helm cli flag `-f` does. This solves #9.

### DIFF
--- a/helm/resource_chart.go
+++ b/helm/resource_chart.go
@@ -9,8 +9,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/ghodss/yaml"
 	"github.com/hashicorp/terraform/helper/schema"
-	"gopkg.in/yaml.v1"
 	"k8s.io/helm/pkg/chartutil"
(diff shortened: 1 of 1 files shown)


mcuadros:
I prefer `values` and `set` as you suggested

cjstone:
Would this be the way to do this with the patch?

```
resource "helm_chart" "jupyterhub" {
    name = "jupyterhub",
    chart = "helm-chart",
    raw_values = "config.yaml"
}
```
config.yaml:

```
hub:
  cookieSecret: "blah"
token:
  proxy: "blah"
singleuser:
  image:
    name: us.gcr.io/blah/blah
    tag: v2.2
```
Doing so produces this:

```
Error applying plan:

1 error(s) occurred:

* helm_chart.jupyterhub: 1 error(s) occurred:

* helm_chart.jupyterhub: error unmarshaling JSON: json: cannot unmarshal string into Go value of type map[string]interface {}
```

Thanks!

PR #31, opened 2017-11-15
upgrade helm to 2.7.x and k8s to 1.8.0

mstrzele:
yep! I'm still using this provider and it would be nice to use the newest helm with it 😉

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -7,12 +7,11 @@ import:
   - plugin
   - terraform
 - package: k8s.io/helm
-  version: ~2.6.0
+  version: ~2.7.0
 - package: gopkg.in/yaml.v1
 
-# hacks for kubernetes v1.7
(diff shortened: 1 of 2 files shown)


hpio:
Any chance we can have this merged please?

Mistobaan:
I would like to have 2.7 helm upgrade, but not the k8s 1.8 upgrade ...

mcuadros:
Fixed by: https://github.com/mcuadros/terraform-provider-helm/commit/a7ff99e4016222e0e8d90998c3e466a797b8adbd

I didn't merge this PR before, because for update the depdencies is not, just update the manifest, also requires to update the `vendor/` folder, and ... since k8s and terrraform are a dependency hell, takes time. But thanks for the PR.

@Mistobaan the update of the client to latest version, doesn't force the usage of 1.8 or 1.9 kubernetes.

Issue #26, opened 2017-09-21
Terraform wants to recreate already existing releases


sagikazarmark:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
Terraform v0.10.4

### Affected Resource(s)
Please list the resources as a list, for example:
- helm_chart

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl
resource "kubernetes_namespace" "logging" {
    metadata {
        name = "logging"
    }
}

resource "helm_chart" "logcollector" {
    name       = "logcollector"
    namespace  = "${kubernetes_namespace.logging.metadata.0.name}"
    chart      = "fluent-bit"
    repository = "${helm_repository.web.name}"
    version    = "0.1.0"
    timeout    = 600

    set {
        name  = "image.fluent_bit.tag"
        value = "0.12.2"
    }

    set {
        name  = "filter.merge"
        value = "On"
    }

    set {
        name  = "backend.type"
        value = "es"
    }

    set {
        name  = "backend.es.host"
        value = "${helm_chart.es.name}-elasticsearch.${kubernetes_namespace.persistence.metadata.0.name}.svc"
    }

    set {
        name  = "resources.limits.memory"
        value = "300Mi"
    }

    set {
        name  = "rbac.create"
        value = true
    }

    # TODO: increase resources
}
```

Happens with several different configurations as well (note that this is only a partial configuration missing the elasticsearch setup)

### Panic Output
None

### Expected Behavior
Terraform should just work

### Actual Behavior
Terraform wants to recreate already existing releases:

```
  + module.dev-k8s.helm_chart.logcollector
      id:                   <computed>
      chart:                "fluent-bit"
      keyring:              "/Users/Y9CRPE/.gnupg/pubring.gpg"
      metadata.#:           <computed>
      name:                 "logcollector"
      namespace:            "logging"
      repository:           "web"
      set.#:                "6"
      set.121059550.name:   "rbac.create"
      set.121059550.value:  "1"
      set.2276311802.name:  "filter.merge"
      set.2276311802.value: "On"
      set.3063255997.name:  "image.fluent_bit.tag"
      set.3063255997.value: "0.12.2"
      set.3604222029.name:  "backend.es.host"
[...]

Issue #32, opened 2017-11-28
New resource shouldn't be required when a chart is updated


mstrzele:
```
-/+ helm_chart.fluent_bit (new resource required)
      id:         "fluent-bit" => <computed> (forces new resource)
      chart:      "http://mstrzele-charts.s3-website-eu-west-1.amazonaws.com/fluent-bit-0.1.0.tgz" => "http://mstrzele-charts.s3-website-eu-west-1.amazonaws.com/fluent-bit-0.1.1.tgz" (forces new resource)
```

As you can see, I've changed the chart version only. IMHO it shouldn't cause a recreation of the whole release.

sagikazarmark:
This looks like a duplicate of #26

mcuadros:
In theory fixed by https://github.com/mcuadros/terraform-provider-helm/pull/39, if not reopen the issue

sagikazarmark:
I started using this provider again for a few weeks now, so far nothing like that.

oba11:
Seem like this issue somehow resurfaced again

PR #39, opened 2018-01-28
add reuse option, add improve create logic

Mistobaan:
This PR tries to fix issues #26 , #32

### DIFF
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -8,10 +8,12 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
+
+	"google.golang.org/grpc"
 
 	"github.com/ghodss/yaml"
 	"github.com/hashicorp/terraform/helper/schema"
-	"google.golang.org/grpc"
 	"k8s.io/helm/pkg/chartutil"
 	"k8s.io/helm/pkg/downloader"
 	"k8s.io/helm/pkg/getter"
@@ -113,6 +115,12 @@ func resourceRelease() *schema.Resource {
 				Default:     false,
 				Description: "Force resource update through delete/recreate if needed.",
 			},
+			"reuse": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
+				Description: "Instruct Tiller to re-use an existing name.",
+			},
 			"recreate_pods": {
 				Type:        schema.TypeBool,
 				Optional:    true,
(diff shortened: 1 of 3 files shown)


helm/resource_release.go
```diff
@@ -382,19 +442,28 @@ var all = []release.Status_Code{
...
-func getRelease(client helm.Interface, d *schema.ResourceData) (*release.Release, error) {
-	name := d.Get("name").(string)
-
+func getRelease(client helm.Interface, name string) (*release.Release, error) {
 	res, err := client.ReleaseContent(name)
 	if err != nil {
-		msg := grpc.ErrorDesc(err)
-		if strings.Contains(msg, "not found") {
-			return nil, ErrReleaseNotFound
+		if status, ok := status.FromError(err); ok {
```
mcuadros: why two this alternative method?

helm/resource_release.go
```diff
@@ -321,17 +381,17 @@ func getChart(d *schema.ResourceData, m *Meta) (c *chart.Chart, path string, err
 		d.Get("keyring").(string),
 	)
 	if err != nil {
-		return
+		return c, path, err
```
mcuadros: not needed, since the return args are named

helm/resource_release.go
```diff
@@ -164,20 +178,54 @@ func resourceReleaseCreate(d *schema.ResourceData, meta interface{}) error {
 	if err != nil {
 		return err
 	}
+	name := d.Get("name").(string)
+out:
+	for {
```
mcuadros: maybe you can move this logic to a function

mcuadros (changes requested):

mcuadros:
Nice PR, but for the next time will be better if you can send several PR or at least several commits.

Mistobaan:
Will break it in a nicer way. This was just a quick show of the changes to comment.

helm/resource_release.go
Mistobaan: yes. Double check if the logic makes sense. Especially if the state of the chart is unknown, should we abort, wait or try to install the new chart anyway?

mcuadros (approved):

sagikazarmark:
Can we expect a release in the near future?

sagikazarmark:
Ping @mcuadros 

This PR fixed one of the two major issues with this provider, would be nice to see a release.

mcuadros:
@sagikazarmark https://github.com/mcuadros/terraform-provider-helm/releases/tag/v0.5.0

sagikazarmark:
Awesome, thanks

PR #47, opened 2018-02-13
resource helm_release doesn't support import

mcanevet:


jhoblitt:
I discovered that import didn't work the hardway.  I think it would be better if there was a cautionary blurb in the docs mention that import doesn't work.

mcanevet:
@jhoblitt what didn't work? Is this unfixable?

jhoblitt:
@mcanevet Trying to import an chart that was deployed manually.

jhoblitt:
Specifically, importing of a manually deployed helm resourced failed and it had to be removed by hand in order to allow the tf `helm_release` resource converge.

burdiyan:
BTW, The provider will try to import a release if it's already deployed, but not present in Terraform state when you apply. So you can try to apply a release as if it were a new one, and it will be imported to the state file. Just make sure the configuration in your Terraform files match what is actually deployed to avoid any problems.

fatmcgav:
I suspect the resource just needs an update to work with the latest Import functionality in Terraform... 

https://github.com/mcuadros/terraform-provider-helm/blob/master/helm/resource_release.go#L31
```go
...
		Importer: &schema.ResourceImporter{
			State: schema.ImportStatePassthrough,
		}
...
```

emmekappa:
I'm having the same issue

lemaral:
@burdiyan I tried to apply on an existing release but got rpc error: code = Unknown desc = a release named <existing release> already exists

meyskens:
Closing in favor for #150

PR #65, opened 2018-04-19
Added a local_files options to mirror -f/--values command

stefanthorpe:
The current values setting requires the use of file() this interpolates before the terraform plan phase. 
This forces the inclusion of values files within the terraform repo. 
Means a separation of values from the code repo, or duplication. Both of which are messy.
By passing file path strings, the files don't get read until executed.

### DIFF
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -3,6 +3,7 @@ package helm
 import (
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"log"
 	"net/url"
 	"os"
@@ -67,6 +68,12 @@ func resourceRelease() *schema.Resource {
 				Type:        schema.TypeList,
 				Optional:    true,
 				Description: "List of values in raw yaml file to pass to helm.",
+				Elem:        &schema.Schema{Type: schema.TypeString},
+			},
+			"value_files": {
+				Type:        schema.TypeList,
+				Optional:    true,
+				Description: "List of values file paths to pass to helm.",
 				Elem:        &schema.Schema{Type: schema.TypeString},
 			},
 			"set": {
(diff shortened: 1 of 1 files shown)


helm/resource_release.go
```diff
@@ -480,6 +487,29 @@ func mergeValues(dest map[string]interface{}, src map[string]interface{}) map[st
...
 	base := map[string]interface{}{}
 
+	// User specified a values files simialar to a -f/--values
+	for _, filePath := range d.Get("value_files").([]interface{}) {
+		currentMap := map[string]interface{}{}
+
+		var bytes []byte
+		var err error
+		if strings.TrimSpace(filePath.(string)) == "-" {
+			bytes, err = ioutil.ReadAll(os.Stdin)
```
meyskens, line 497: I don't think this would work by the way terraform providers are loaded

legal90:
I'm not a maintainer of this provider, but I'm really worrying about how it is developed. So I would like to add my 5 cents. 

I don't think that is a good idea to have such an option for a local file path.

As I see the "Terraform ideology" is that it tracks the state of the resource, matches it with the configuration and updates the resource according to the configuration changes.
Having an attribute which stores just the local path to values file doesn't really makes sense. The state should store the real input which affects the resource state and we already have a plenty of attributes for that: `values`, `set`, `set_string`, `set_secret`.

> Means a separation of values from the code repo, or duplication. Both of which are messy.

To avoid duplication you can specify more than one item for `values`, for example:
```hcl
values = [
  "${file("${path.module}/../shared_values/my_chart.yaml")}",
  "${file("${path.module}/values/my_chart.yaml")}"
]
```
Here `./values/my_chart.yaml` will be merged on top of the `../shared_values/my_chart.yaml`. That's how you can split your values and share/reuse peaces by multiple resources, if that's the case.

rporres:
As stated by @legal90 and discussed in https://github.com/terraform-providers/terraform-provider-helm/issues/66, this is not something we want at the moment. Thanks for your contribution, anyway.

PR #74, opened 2018-05-14
fixes issue/64: Add helms set-string option (set_string)

stevenaldinger:
Allows using `set_string` the same way as `set` which translates into helm's `set-string` and allows using characters like commas and periods.

Usage:

```
    set_string {
        name = "image.tags"
        value = "registry\\.io/terraform-provider-helm\\,example\\.io/terraform-provider-helm"
    }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,11 @@ resource "helm_release" "my_database" {
         name = "mariadbPassword"
         value = "qux"
     }
+
+    set_string {
+        name = "image.tags"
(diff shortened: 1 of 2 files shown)


stevenaldinger:
I've updated this on my fork's master to keep it from getting stale but if there's some issue / reason this isn't being merged in if you let me know I'd fix it vs. trying to maintain a forked version that actually works

PR #85, opened 2018-06-25
Adding wait option to install and upgrade

dstori:
Currently `wait` option is fixed (`true`) both to install and upgrade. 
This PR is to add support to this flag in the terraform template.


sebastien-prudhomme:
@mcuadros I've seen that you don't want a nowait option as this the way Terraform is working.

I'm using your provider a lot and i just can say that adjusting the timeout option is really annoying and not a good option for me.

I will be a really happier user if you change your opinion about adding a feature that allow to disable the wait option.

meyskens (approved):

PR #89, opened 2018-07-13
Require explicit config_path and allow installTiller from on the fly certs

geriBatai:
This PR fixes two issues for us. 

1. It no longer allows implicit `config_path` setting. This was a problem when the same terraform config was applied by different people, and terraform run had different results based on their environment setup.

2. It allows us to pass certificate data into helm out of other terraform resources (e.g. `tls_self_signed_certificate`) and deploy these certificates as secrets for tiller (helm installer requires files on a filesystem, that's why we're pushing them through `tempFileOrPath`) .

### DIFF
--- a/helm/provider.go
+++ b/helm/provider.go
@@ -4,7 +4,9 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
+	"io/ioutil"
 	"log"
+	"os"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


geriBatai:
@mcuadros any comment?

meyskens:
@geriBatai can you rebase this on master so we can run tests? Thanks!

aaronmell:
This can be closed, I've fixed this in V3 in a similar manner, and Tiller has been removed, so that part isn't needed.

dak1n1:
Closing since we're no longer doing updates to Helm 2 versions of the provider.

PR #90, opened 2018-07-14
Read namespace and version keys from release to ensure they are correct

cezarsa:
This commit changes the behavior of diffs related to version and namespace changes. The individual changes are:

* The Read operation on the release resource updates `version` and `namespace` keys;
* The `version` key is now a Computed value;
* A customized diff function tries to read the new `version` for the provided chart during diff generation, possibly triggering a change if the version was implicitely changed.

These changes combined make the release resource able to detect changes to the chart version even if they happened outside terraform (e.g. manual helm upgrade call) or if the version was not declared and the chart points to a directory.

This commit also includes a test that demonstrate the new behavior by applying a release using a chart from a directory and updating its version.

### DIFF
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -29,11 +29,12 @@ var ErrReleaseNotFound = errors.New("release not found")
 
 func resourceRelease() *schema.Resource {
 	return &schema.Resource{
-		Create: resourceReleaseCreate,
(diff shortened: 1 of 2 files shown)


mcuadros:
Awesome, thanks for the contribution

smola (approved):

PR #91, opened 2018-07-14
Update README

costimuraru:
Tried to run the example from the README with `name = my_datasase` and got the following error:

```
Secret "my_datasase-mariadb" is invalid: metadata.name: Invalid value: "my_datasase-mariadb": a DNS-1123 subdomain must consist of lower case alphanumeric characters
```

Also, while quickly following the installation guide on the main page, I fetched the Linux package though I have MacOS and was wondering I got an error on `terraform plan`

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,12 +33,19 @@ distributions from the [Releases](https://github.com/mcuadros/terraform-provider
 
 Download and uncompress the latest release for your OS. This example uses the linux binary.
 
+#### Linux:
 ```sh
 > wget https://github.com/mcuadros/terraform-provider-helm/releases/download/v0.5.1/terraform-provider-helm_v0.5.1_linux_amd64.tar.gz
 > tar -xvf terraform-provider-helm*.tar.gz
 ```
 
-Now copy the binary to the Terraform's plugins folder, if is your first plugin maybe isn't present.
+#### MacOS:
+```sh
+> wget https://github.com/mcuadros/terraform-provider-helm/releases/download/v0.5.1/terraform-provider-helm_v0.5.1_darwin_amd64.tar.gz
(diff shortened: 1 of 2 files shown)


docs/README.md
```diff
@@ -11,7 +11,7 @@ The Helm provider is used to deploy software packages in Kubernetes. The provide
 
 ```hcl
 resource "helm_release" "my_database" {
-    name      = "my_datasase"
+    name      = "mydatasase"
```
Jeeppler, line 14: Should this not be `mydatabase`?

README.md
```diff
@@ -33,12 +33,19 @@ distributions from the [Releases](https://github.com/mcuadros/terraform-provider
...
 ```
 
-Now copy the binary to the Terraform's plugins folder, if is your first plugin maybe isn't present.
+#### MacOS:
+```sh
+> wget https://github.com/mcuadros/terraform-provider-helm/releases/download/v0.5.1/terraform-provider-helm_v0.5.1_darwin_amd64.tar.gz
+> tar -xvf terraform-provider-helm*.tar.gz
+```
+
+Now copy the binary to the Terraform's plugins folder. If this is your first plugin the folder might be present.
```
Jeeppler, line 48: How about adding a heading which says installing plugin into Terraform? This way it would be clearer, that the install processes is platform neutral (at least for *nix).

```
#### Install Plugin

Now copy the binary to the Terraform's plugins folder. If this is your first plugin the folder might be present.
```

Alternative you could add just dashes to separate this section from the MacOS text.

meyskens:
@costimuraru Thanks for your PR. I seems in meantime the example has been corrected, regarding the install instructions: since we're now official these will change as `terraform init` will install this provider soon.

Issue #107, opened 2018-09-01
Upgrade dependencies to helm 2.10


emmekappa:
Helm 2.10 is out

mcuadros:
https://github.com/mcuadros/terraform-provider-helm/releases/tag/v0.6.0

PR #108, opened 2018-09-01
Updating helm to 2.10

emmekappa:
Related issue: #107


smola (changes requested):
This would require upgrading `vendor` with `dep ensure`. Otherwise, the change on `Gopkg.toml` has no effect.

mcuadros:
The process of update dependencies, is a complex task, I am taking care of it now. 

Thanks for the PR.

PR #123, opened 2018-10-05
recreate resources when chart values change

niroliyanage:
Awesome project and is heavily used, this will give me the ability to recreate resources when chart values change

### DIFF
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -68,6 +68,7 @@ func resourceRelease() *schema.Resource {
 			"values": {
 				Type:        schema.TypeList,
 				Optional:    true,
+				ForceNew:    true,
 				Description: "List of values in raw yaml file to pass to helm.",
 				Elem:        &schema.Schema{Type: schema.TypeString},
 			},
@@ -198,6 +199,7 @@ func resourceRelease() *schema.Resource {
 						"values": {
 							Type:        schema.TypeString,
 							Computed:    true,
+							ForceNew:    true,
 							Description: "The raw yaml values used for the chart.",
 						},
 					},


helm/resource_release.go
```diff
@@ -198,6 +199,7 @@ func resourceRelease() *schema.Resource {
 						"values": {
 							Type:        schema.TypeString,
 							Computed:    true,
+							ForceNew:    true,
```
meyskens, line 202: The values here are computed, so `ForceNew` is of no use here

meyskens:
Thanks for your PR! I don't really see why `ForceNew` is needed for new values? `ForceNew` will delete the release and recreate here, in this case I think an `helm upgrade` should be enough for a values change. 
Would love to hear your reasoning on this.

meyskens (changes requested):
Thanks for your PR! I don't really see why `ForceNew` is needed for new values? `ForceNew` will delete the release and recreate here, in this case I think an `helm upgrade` should be enough for a values change.
Would love to hear your reasoning on this.

niroliyanage:
There are instances where we dont manage the helm charts. there are a few use cases for us that  we only need to change the values forcing the recreation of the resource using the new values automatically. We have another CI platform maintaining the change orchestration. If this defeats the purpose i can get around this by maintaining our own charts repo.

ATM, adding new values would essentially compute the new values and modifies the resource and then manual intervention is required to update it

rporres:
@niroliyanage We cannot force a new resource just because we're changing values. That would mean outage for already deployed releases where we just want a rolling upgrade.

If you want that behaviour, you can just change whatever you need and taint your resource to force recreation, but it should not be the default way of doing things.

PR #124, opened 2018-10-05
Statically compile provider

tecnobrat:
The official docker image for terraform is based on alpine linux, which doesn't have glibc installed.  So we need a statically compiled version.


hairmare:
It looks like #111 will do the same thing on non travis builds as well.

mcuadros:
Duplicated https://github.com/mcuadros/terraform-provider-helm/pull/111 but thanks

PR #128, opened 2018-10-16
Always update local healm repositories when reading resources

divoxx:
The local repository is only created when `terraform apply` is run. This makes impossible for someone on a different machine to run a `terraform plan` since the repository will be missing.

In general, local config resources should be consistently applied as needed since their state are not persisted on the infrastructure. That said, I'm not sure if this is something that was brought up in the past or not.

Cheers!


meyskens (approved):
I have had exactly the same issue! Looks good!

PR #131, opened 2018-10-17
Fix another documentation link

meyskens:


rporres:
LGTM, but travis is failing. Totally unrelated.

rporres (approved):

meyskens:
@rporres Can't find why travis is breaking, i did notice that we should update some stuff in travis but it will be replaced by the hashicorp ci soon.

PR #137, opened 2018-10-23
Fix recursion caused by auto insert of repo

divoxx:
My previous change on 36428fcffb9355a16db4af7d601d0b52f379b3d0 is causing a recursion between the read and create methods of the provider.

This uses the Exists hook in order to create the resource when necessary without causing the recursion.

### DIFF
--- a/helm/resource_repository.go
+++ b/helm/resource_repository.go
@@ -19,6 +19,7 @@ func resourceRepository() *schema.Resource {
 		Create: resourceRepositoryCreate,
 		Read:   resourceRepositoryRead,
 		Delete: resourceRepositoryDelete,
+		Exists: resourceRepositoryExists,
 		Schema: map[string]*schema.Schema{
 			"name": {
 				Type:        schema.TypeString,
@@ -113,11 +114,6 @@ func resourceRepositoryCreate(d *schema.ResourceData, meta interface{}) error {
 func resourceRepositoryRead(d *schema.ResourceData, meta interface{}) error {
 	m := meta.(*Meta)
 
-	err := resourceRepositoryCreate(d, m)
-	if err != nil {
-		return err
-	}
-
 	r, err := getRepository(d, m)
 	if err != nil {
 		return err
@@ -137,6 +133,15 @@ func resourceRepositoryDelete(d *schema.ResourceData, meta interface{}) error {
 	debug("%q has been removed from your repositories\n", name)
 	d.SetId("")
 	return nil
+}
+
+func resourceRepositoryExists(d *schema.ResourceData, meta interface{}) (bool, error) {
+	err := resourceRepositoryCreate(d, meta)
+	if err != nil {
+		return false, err
+	}
+
+	return true, nil
 }
 
 func setIDAndMetadataFromRepository(d *schema.ResourceData, r *repo.Entry) error {
--- a/main.go
+++ b/main.go
@@ -1,8 +1,8 @@
 package main
 
 import (
+	"github.com/divoxx/terraform-provider-helm/helm"
 	"github.com/hashicorp/terraform/plugin"
-	"github.com/terraform-providers/terraform-provider-helm/helm"
 )
 
 func main() {
--- a/main.go
+++ b/main.go
@@ -1,8 +1,8 @@
 package main
 
 import (
-	"github.com/divoxx/terraform-provider-helm/helm"
 	"github.com/hashicorp/terraform/plugin"
+	"github.com/terraform-providers/terraform-provider-helm/helm"
 )
 
 func main() {


main.go
```diff
@@ -1,8 +1,8 @@
 package main
 
 import (
+	"github.com/divoxx/terraform-provider-helm/helm"
```
rporres: This import should not be changed

rporres:
Could you please provide provide a way to reproduce the problem you're solving?

main.go
divoxx: I needed to change it so that the correct package gets compiled when testing, so it's correct in the context of the `divoxx/terraform-provider-helm` repo (we are using that branch).

My thought was to change it back when/if this is ready to be merged. Does that work?

divoxx:
@rporres simply running `terraform plan` when there is a repository definition will cause it to hang due the recursion introduced in https://github.com/terraform-providers/terraform-provider-helm/pull/128.

I'm not sure how we didn't catch this earlier. Also, I'm not sure what would be a good strategy to test this, but I'm happy to add tests if you can provide some directions.

main.go
rporres: Hmm, you can clone your fork under `$GOPATH/src/github.com/terraform-providers/terraform-provider-helm` so that you don't need to do this

divoxx: Yeah, this would work in this case since this project is non go-gettable. It's just force of habit when working on golang forks, to make sure the repo is still go-gettable and correctly loads the packages from the repo instead of pulling from the original.

I'll update.

helm/resource_repository.go
```diff
@@ -139,6 +135,15 @@ func resourceRepositoryDelete(d *schema.ResourceData, meta interface{}) error {
 	return nil
 }
 
+func resourceRepositoryExists(d *schema.ResourceData, meta interface{}) (bool, error) {
+	err := resourceRepositoryCreate(d, meta)
```
rporres, line 139: Terraform documentation https://godoc.org/github.com/hashicorp/terraform/helper/schema#Resource  recommends not to modify the *ResourceData passed to `Exists` so I'm not convinced about this approach.

From the documentation, it should recreate the resource if you look for it and return false if it doesn't exist. Have you tried that?

divoxx: That wouldn't fix the original problem.

The first PR was an attempt to fix a problem related to repositories not being correctly added to the local repository list, which would cause `terraform plan` (or any other read only command) to fail on a different setup than the one that ran `terraform apply`.

A way to reproduce the original issue is to have a `helm_repository` setup, then run `terraform apply`, and then on another computer try to run `terraform plan` or `terraform refresh` which would raise an error because the repository wouldn't exist on the local repositories.yml.

I understand that this approach is weird since Exists is only supposed to return whether the resource exists, but a helm repository is not a "remote" resource that is created and persisted in the infrastructure. It's a local resource and it needs to somehow be enforced to exist on the client side.

Since only Exists or Read are called when read only commands are run (like `plan` or `refresh`), it needs to be enforced within one of those. The first attempt was using `Read` but calling `Create` from it causes a recursion since `Create` also calls `Read`.

rporres: I've made the exact test that you're mentioning using a provider where I've added this `Exists` function:
```go
func resourceRepositoryExists(d *schema.ResourceData, meta interface{}) (bool, error) {
	m := meta.(*Meta)

	_, err := getRepository(d, m)
	if err == ErrRepositoryNotFound {
		return false, nil
	} else {
		return true, err
	}

}
```
and whenever it is run in the second computer the plan adds the creation of the local repo that was not there... 

Anyway the real problem is the `helm_repository` resource that most likely should be deprecated and make `helm_release` download charts.

divoxx: I'm not sure I understand. 

My understanding is that `helm_repository` is not making changes to the infrastructure/state, instead it's simply adding to the local `repositories.yaml` file. In which case, according to your snippet, `terraform plan` or `terraform refresh` would always have different results, depending on the computer that runs it and whether it has been ran before.

This problem is a lot more apparent when you have terraform commands being executed inside containers, for example in CI or projects using docker for uniform development environments.

In general, changes to local configurations that are not tracked by the state should be idempotent and behave the same across all computers, and not have varying plans of execution depending on the local state not tracked by terraform.

On the other hand, if the plan is to deprecate `helm_repository`, then that's a separate discussion which I don't have a lot of context. The fact that my previous PR got merged by another contributor led me to think that this is a problem worth fixing, since it is pretty disruptive and annoying if there are a lot of devs working on the same project or if you're using terraform within containers or any other immutable infrastructure. 

This specific PR is only fixing the infinite recursion problem introduced by the [previous PR](https://github.com/terraform-providers/terraform-provider-helm/pull/128). If the overall problem is not something that needs to be fixed, then feel free to revert the latter and we can maintain our own fork.

rporres: Of course the problem needs to be addressed, we just have different ideas about how to fix it :)

The problem with your approach is that it is doing things that are not recommended by the `terraform` documentation. Furthermore `terraform plan` will be making changes and that must not happen.

I understand that it does not look right that runs in different servers have different plans but it is the way `helm_repository` works as it manages a local resource. At the bottom of the problem is that the resource is flawed for that very same reason.

Following the spirit of your last paragraph, I will close this PR, revert the previous PR and then create a new one with the solution I've proposed as the issue still needs solution. Then I will take a further look to the `helm_repository` resource as it should change to avoid the issues we're experiencing.

I would really like to thank you for your contribution and the time you've spent discussing this issue.

PR #141, opened 2018-10-25
Use helm 2.11.0

meyskens:
Needs some acceptance testing but should not cause much trouble


mcuadros:
Doesn't compile properly

meyskens:
@mcuadros will take a look, it did on my laptop so it probably didn't push something :disappointed:

meyskens:
does not work at all, will retry again

meyskens:
Hopefully #143 works

PR #142, opened 2018-10-26
Bump Go version to 1.10.4 in Travis CI

alexsomesan:
This aligns the Go version in Travis to the one used for releases.

@meyskens @rporres @terraform-providers/ecosystem

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ services:
 - docker
 language: go
 go:
-- 1.8.1
+- 1.10.4
 
 install:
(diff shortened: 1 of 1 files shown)


bflad:
Any reason to not go to `1.11.1`?

alexsomesan:
None other than TC release environment being on 1.10.x and wanted to keep things consistent.

1.11.x behaves subtly different to 1.10.x even in GOPATH mode (as I found out the hard way with K8S provider).

rporres (approved):

PR #153, opened 2018-11-12
add the ability to use sensitive values

### LINKED ISSUE #80: Add ability to hide sensitive values
This is an enhancement request.

Add an additional property to `set` for `helm_release` resources that allow the value to be marked as 'sensitive' and not displayed in the state diff during a plan or apply.

For example:

```hcl
resource "helm_release" "my_database" {
    name      = "my_datasase"
    chart     = "stable/mariadb"

    set {
        name  = "mariadbUser"
        value = "foo"
    }

    set {
        name      = "mariadbPassword"
        value     = "THIS IS SUPER SECRET"
        sensitive = true
    }
}
```

### Terraform Version

```
terraform version
Terraform v0.11.7
+ provider.helm (unversioned)
```

Note: though `provider.helm` is said to be `unversioned`, it has version `v0.5.1`.

### Affected Resource(s)

Any `helm_release` resource is affected by this.

```hcl
resource "helm_release" "my_database" {
    name      = "my_datasase"
    chart     = "stable/mariadb"

    set {
        name  = "mariadbUser"
        value = "foo"
    }

    set {
        name = "mariadbPassword"
        value = "THIS IS SUPER SECRET"
    }
}
```

### Expected Behavior

Secret values should not be displayed. For example, when using the aws provider, some resources (eg aws_rds_instance) have values that are masked with `<sensitive>`.

### Actual Behavior

Secrets are displayed plain text:

```
  + helm_release.my_database
      id:                               <computed>
      chart:                            "stable/mariadb"
      disable_webhooks:                 "false"
      force_update:                     "false"
      metadata.#:                       <computed>
      name:                             "my_datasase"
      namespace:                        "default"
      recreate_pods:                    "false"
      reuse:                            "false"
      reuse_values:                     "false"
      set.#:                            "2"
      set.1144798963.name:              "mariadbUser"
      set.1144798963.value:             "foo"
      set.1144798963.name:              "mariadbPassword"
      set.1144798963.value:             "THIS IS SUPER SECRET"
      timeout:                          "300"
      verify:                           "false"
```

### Steps to Reproduce

1. Ensure terraform-provider-helm is installed
2. Ensure kubectl and helm are installed and configured
3. `terraform apply` | `terraform plan`

rmb938:
Fixes #80 

I wanted to do this via a DiffSuppressFunc with a sensitive bool instead of creating a whole new value but I couldn't figure out a good way to make it work.

### DIFF
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -3,7 +3,6 @@ package helm
 import (
 	"errors"
 	"fmt"
-	"log"
 	"net/url"
 	"os"
 	"path/filepath"
@@ -84,6 +83,24 @@ func resourceRelease() *schema.Resource {
 						"value": {
 							Type:     schema.TypeString,
 							Required: true,
+						},
+					},
+				},
+			},
+			"sensitive_set": {
+				Type:        schema.TypeSet,
+				Optional:    true,
+				Description: "Custom sensitive values to be merge with the values.",
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"name": {
+							Type:     schema.TypeString,
+							Required: true,
+						},
+						"value": {
+							Type:      schema.TypeString,
+							Required:  true,
+							Sensitive: true,
 						},
 					},
 				},
@@ -544,6 +561,17 @@ func getValues(d *schema.ResourceData) ([]byte, error) {
 		}
 	}
 
+	for _, raw := range d.Get("sensitive_set").(*schema.Set).List() {
+		set := raw.(map[string]interface{})
+
+		name := set["name"].(string)
+		value := set["value"].(string)
+
+		if err := strvals.ParseInto(fmt.Sprintf("%s=%s", name, value), base); err != nil {
+			return nil, fmt.Errorf("failed parsing key %q with sensitive value, %s", name, err)
+		}
+	}
+
 	for _, raw := range d.Get("set_string").(*schema.Set).List() {
 		set := raw.(map[string]interface{})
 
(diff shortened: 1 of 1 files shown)


helm/resource_release.go
```diff
@@ -555,12 +583,7 @@ func getValues(d *schema.ResourceData) ([]byte, error) {
 		}
 	}
 
-	yaml, err := yaml.Marshal(base)
-	if err == nil {
-		log.Printf("---[ values.yaml ]-----------------------------------\n%s\n", yaml)
```
rmb938, line 560: removing this to prevent logging of secrets

helm/resource_release.go
```diff
@@ -88,6 +87,24 @@ func resourceRelease() *schema.Resource {
 					},
 				},
 			},
+			"sensitive_set": {
```
meyskens: To stay in the line of what we have already `set` `set_string` maybe this should be `set_sensitive`?

rmb938: Ah yes. I can get that changed.

rmb938: done

legal90:
Sorry for interrupting, but I believe it could be quite confusing to have a new option `set_sensitive`, since unlikely `set` and `set_string`, this one does not map to the corresponding argument of `helm` CLI.

Did you look at the possibility to implement it via `sensitive` boolean flag for `set` map, as was proposed in #80? 
I was going to look on it pretty soon.

rmb938:
Well if we set the `senstive` flag then all set values will be marked as sensitive so then a plan will be pretty useless for non-sensitive values. 

I also tried using a Diff supressor but I couldn't really figure out a way to make it work nicely. I am open to ideas though :)

meyskens:
I took a look into that idea too, it is not possible to set that unless you use the DiffSupressor (which I found out about today), not exactly sure how that would work but it could work I think

rmb938:
I tried using the DiffSupressor and a `senstive` boolean in `set` and I couldn't really find a good way to make it work. It doesn't seem well suited for complex datasets.

rmb938:
@meyskens Any updates?

rmb938:
@meyskens Can we look into getting this merged?

meyskens:
@rmb938 Sorry for the delay, was still doubting about the removing of the logging but will PR a fix for that later.

Issue #79, opened 2018-06-04
Show the diff manifest in plan.


mcanevet:
Unless I'm wrong, this provider will not detect changes if neither the chart version, nor the values changed. However, one can have made manual changes using the k8s api (update or delete objects), and this provider should detect this and reconcile using what is described in the Terraform manifest.

I think we can achieve this using the helm diff plugin, we could store the current state in a computed field and then, find a way to compare it to what it will deploy.

mcanevet:
I tried to implement this last week end but it is not that easy. After discussing about that on #committer-terraform on HashCorp's slack, I had some inputs from @apparentlymart. What we need is:
- a `manifest` computed argument. This part is easy, I'll probably prepare a PR for that.
- Launch a dry-run of the release to refresh the `manifest` argument in the `CustomizeDIff` function. But this may have some side effects.

mcanevet:
I copy/paste here the discussion I just had on Slack regarding this issue:
```
Hello,
I need some advise regarding a contribution I'd like to make to the helm_release resource of the helm provider. The "version" argument is "optional", "computed". If one does not specify it, the latest version of the helm chart is taken on creation. When a new version is released, I would expect that terraform tries to upgrade the release to the latest version, but it is not the case because Terraform does not know that a new version is available.
I know how to fetch the latest version by doing a dry-un upgrade of a helm release, but what I don't really know is in what part of the code I should populate the new value of the "version" argument so that terraform can propose an upgrade.

rileykarson [5:32 PM]
We use datasources in the Google provider so that Terraform knows an upgrade is available in similar cases, if it helps: https://www.terraform.io/docs/providers/google/d/google_container_engine_versions.html (edited)
Terraform by HashiCorp
Google: google_container_engine_versions - Terraform by HashiCorp
Provides lists of available Google Container Engine versions for masters and nodes.

mcanevet [5:39 PM]
hmmm ok. That would do the trick for this use case. But I have another contribution in mind that would consist of adding a "manifest" computed argument to store the generated manifest. This would allow to show a clean difference of what would be really deployed in a k8s cluster with the new diff renderer of Terraform v0.12 (https://github.com/hashicorp/terraform/issues/15180#issuecomment-435241641). But for this use case I need to populate the "manifest" argument with a dry-run

apparentlymart [5:40 PM]
[...]

PR #154, opened 2018-11-13
Add manifest attribute

mcanevet:
This is a first step to implement #79.
Next step would be to refresh the manifest launching a dry-run so that we can have a clean output of what will change just like using the helm diff plugin

### DIFF
--- a/website/docs/release.html.markdown
+++ b/website/docs/release.html.markdown
@@ -57,6 +57,7 @@ The `set` block supports:
 In addition to the arguments listed above, the following computed attributes are
 exported:
 
+* `release` - The string representation of the rendered template.
 * `metadata` - Block status of the deployed release.
 
 The `metadata` block supports:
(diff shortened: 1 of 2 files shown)


meyskens:
@mcanevet thanks for your contribution. Is it possible to also add some (acceptance) tests in this PR?

website/docs/release.html.markdown
```diff
@@ -57,6 +57,7 @@ The `set` block supports:
 In addition to the arguments listed above, the following computed attributes are
 exported:
 
+* `release` - The string representation of the rendered template.
```
legal90: I guess, it should be `manifest` instead.

meyskens:
Going to close this for now as there is no input being filled in `manifest` which could confuse people who would like to use that output.
I suggest to re-use this commit in a later PR where there is logic added.

PR #166, opened 2018-12-13
Fix acceptance tests

meyskens:
This PR adds fixes for acceptance tests that are failing for reasons not related to the provider


rporres (approved):
I do agree that I don't see any value in enabling persistence to test this provider

PR #172, opened 2018-12-17
Fix build after dependency updates

alexsomesan:
Builds were failing for the OpenBSD target. This change removes OpenBSD from the list of active targets.

As a result `make compile` now behaves as expected.
```
------------------------------------------------------------
~/work/src/github.com/terraform-providers/terraform-provider-helm(fix-build*) » make compile                                                                                                                                                                                                               alex@alexs-macbook
==> Checking that code complies with gofmt requirements...
Number of parallel builds: 7

-->     freebsd/arm: github.com/terraform-providers/terraform-provider-helm
-->       linux/386: github.com/terraform-providers/terraform-provider-helm
-->     freebsd/386: github.com/terraform-providers/terraform-provider-helm
-->     linux/amd64: github.com/terraform-providers/terraform-provider-helm
-->     windows/386: github.com/terraform-providers/terraform-provider-helm
-->    darwin/amd64: github.com/terraform-providers/terraform-provider-helm
-->   windows/amd64: github.com/terraform-providers/terraform-provider-helm
-->   freebsd/amd64: github.com/terraform-providers/terraform-provider-helm
-->       linux/arm: github.com/terraform-providers/terraform-provider-helm
------------------------------------------------------------
```
@terraform-providers/ecosystem 
@meyskens @rporres

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ assets
 .terraform
 _output/
 .vscode
+/results
--- a/scripts/compile.sh
+++ b/scripts/compile.sh
@@ -11,7 +11,7 @@ PROVIDER_NAME="helm"
 TARGET_DIR="$(pwd)/results"
 XC_ARCH=${XC_ARCH:-"386 amd64 arm"}
 XC_OS=${XC_OS:=linux darwin windows freebsd openbsd solaris}
-XC_EXCLUDE_OSARCH="!darwin/arm !darwin/386 !solaris/amd64"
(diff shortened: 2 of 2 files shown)


meyskens:
Thanks! do we have any idea why this is happening?

meyskens (approved):

alexsomesan:
@meyskens Yes, I think I do have an idea. 

It looks like the missing functions it's complaining about (like `unix.IoctlGetTermios` or `unix.IoctlSetTermios`) are missing from the vendored version of `golang.org/x/sys/unix` package.

Newer versions of it seem to have those functions, but since this is an indirect dependency, I'm now experimenting to see if whatever other package that dragged these in will break if I update `golang.org/x/sys/unix`.

PR #182, opened 2019-01-11
Log values.yaml in case of errors and filtering sensitive values

meyskens:
in #153 the debug log for values was removed.
This adds it back in a form where it replaces sensitive values.

### DIFF
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -3,6 +3,7 @@ package helm
 import (
 	"errors"
 	"fmt"
+	"log"
 	"net/url"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


legal90:
@meyskens #150 also includes the introduction of new attribute - `overrides`, which will be saved in state. Maybe that could be used for the debugging purposes too ?

rporres (approved):

rporres:
LGTM 

@legal90 Your comment is more related to #150 than about the code in this very PR, isn't it?

legal90:
@meyskens Ah, you're right. Sorry, I didn't notice that this is a PR, no an issue (feature proposal).
I actually like the implementation of sensitive data suppression here 👍 

And we might need to rework #150 a little bit, after this one is merged.

PR #184, opened 2019-01-16
Add "overrides" attributes accumulating all overridden values for the release

legal90:
_Note:_  #150 will need to be rebased after merging this PR since it re-implements `"overrides"` attribute according to latest changes in master branch. I can handle this later too.

This PR does the following:
- Adds a new computed attribute called `overrides` which accumulates all values from `values`, `set`, `set_secret` and `set_string` attributes merged in the order they mentioned. This new attribute replaces the old `metadata.0.values` since it needs to be updated by `CustomizeDiff` func. 

- Changes the resource update logic. Now the update of the release will happen only if any of `overrides`, `status`, `version` or `reuse_values` attributes are changed or any of `recreate_pods` or `force_update`. Otherwise, there is no reason to call Chart's `UpdateRelease` API, so only the local state will be updated. 
That allows to safely move the same values between `values` / `set` / `set_secret` / `set_string` attributes, add comments & spaces to YAML, etc.  without calling a remote update on the tiller side.

This PR is also important for the future implementation of Importer (like #150), which could just dump all values to `overrides` without worrying about the diff caused by configuration in `*.tf`.

### DIFF
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -262,7 +262,7 @@ func prepareTillerForNewRelease(d *schema.ResourceData, c helm.Interface, name s
 
 		switch r.Info.Status.GetCode() {
 		case release.Status_DEPLOYED:
-			return setIDAndMetadataFromRelease(d, r)
+			return setAttributesFromRelease(d, r)
 		case release.Status_FAILED:
 			// delete and recreate it
 			debug("release %s status is FAILED deleting it", name)
@@ -360,7 +360,7 @@ func resourceReleaseCreate(d *schema.ResourceData, meta interface{}) error {
 		return err
 	}
 
-	return setIDAndMetadataFromRelease(d, res.Release)
+	return setAttributesFromRelease(d, res.Release)
 }
 
 func resourceReleaseRead(d *schema.ResourceData, meta interface{}) error {
@@ -379,10 +379,10 @@ func resourceReleaseRead(d *schema.ResourceData, meta interface{}) error {
 
 	//  d.Set("values_source_detected_md5", d.Get("values_sources_md5"))
 
-	return setIDAndMetadataFromRelease(d, r)
+	return setAttributesFromRelease(d, r)
 }
 
-func setIDAndMetadataFromRelease(d *schema.ResourceData, r *release.Release) error {
+func setAttributesFromRelease(d *schema.ResourceData, r *release.Release) error {
 	d.SetId(r.Name)
 	d.Set("version", r.Chart.Metadata.Version)
 	d.Set("namespace", r.Namespace)
@@ -445,7 +445,7 @@ func resourceReleaseUpdate(d *schema.ResourceData, meta interface{}) error {
 		return err
 	}
 
-	return setIDAndMetadataFromRelease(d, res.Release)
+	return setAttributesFromRelease(d, res.Release)
 }
 
 func resourceReleaseDelete(d *schema.ResourceData, meta interface{}) error {
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -197,6 +197,13 @@ func resourceRelease() *schema.Resource {
 				Computed:    true,
 				Description: "Status of the release.",
 			},
+			"overrides": {
+				Type:        schema.TypeString,
+				Computed:    true,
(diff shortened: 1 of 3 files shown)


helm/resource_release.go
```diff
@@ -189,6 +189,13 @@ func resourceRelease() *schema.Resource {
 				Computed:    true,
 				Description: "Status of the release.",
 			},
+			"overrides": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The result YAML file got by merging all passed `set`, `set_string`, `set_sensitive` and `values`",
+				// Mark as sensitive to avoid printing the diff, as it might contain sensitive data
+				Sensitive: true,
```
legal90, line 205: _Note:_ Without `Sensitive: true` we can have a detailed diff in the plan output, so we can see all exact changes of the result YAML. I like it more than just meaningless `<sensitive> => <sensitive>`  diff

However, since `set_sensitive` has been added, that could reveal such secret values. I don't see any way to avoid it, so I had to mark this attribute ultimately `Sensitive`. Let me know if I should change this.

legal90:
This is how importer could look like after this PR is merged:
https://github.com/legal90/terraform-provider-helm/compare/add-overrides-attr...legal90:add-importer-v2

legal90:
I updated acceptance tests. Added the one verifying that no update is triggered if values (the content of `overrides`) are not changed.

legal90:
Could anyone from maintainers review this, please? 🙏 
/cc: @alexsomesan @meyskens

legal90:
@alexsomesan, @rporres Is there any chance for it to be reviewed & included to v0.8.0? 🙏

rporres:
@legal90 not for 0.8.0, but we'll take a look into this ASAP.

legal90:
Could you guys please review this PR?
I see it as the key step on the way to having an import support for `helm_release`.

helm/resource_release.go
invidian: Without this PR, detailed diff leaks into plan anyway, correct? What about having something like `overrides_sensitive`?

invidian:
I tested this PR together with #150 and it seems, that it's still crashing, when I change the value passed to the resource. I do the following:
```
  values     = [
<<EOF
nodes:
${join("", formatlist("  - '%s'\n", hcloud_server.servers.*.name))}
EOF
  ]
```

And error from Terraform:
```
    Terraform Version: 0.11.13
    Resource ID: helm_release.ceph-cluster
    Mismatch reason: attribute mismatch: overrides
    Diff One (usually from plan): *terraform.InstanceDiff{mu:sync.Mutex{state:0, sema:0x0}, Attributes:map[string]*terraform.ResourceAttrDiff{"overrides":*terraform.ResourceAttrDiff{Old:"nodes:\n- testing-1.xxx\n- testing-2.xxx\n- testing-3.xxx\n", New:"nodes:\n- testing-1.xxx\n- testing-2.xxx\n", NewComputed:false, NewRemoved:false, NewExtra:interface {}(nil), RequiresNew:false, Sensitive:false, Type:0x0}}, Destroy:false, DestroyDeposed:false, DestroyTainted:false, Meta:map[string]interface {}(nil)} 
    Diff Two (usually from apply): *terraform.InstanceDiff{mu:sync.Mutex{state:0, sema:0x0}, Attributes:map[string]*terraform.ResourceAttrDiff(nil), Destroy:false, DestroyDeposed:false, DestroyTainted:false, Meta:map[string]interface {}(nil)}

```

I'll try to investigate.

helm/resource_release.go
legal90: > Without this PR, detailed diff leaks into plan anyway, correct?

Yes, that's correct. But I don't see how `overrides_sensitive` would help in this case.

legal90:
@invidian I think, it's related to the interpolation syntax, not to these PRs. Try using this:
```
values = "nodes: ${jsonencode(hcloud_server.servers.*.name)}"
```

invidian:
>@invidian I think, it's related to the interpolation syntax, not to these PRs. Try using this:

@legal90 unfortunately, I get the exact same error with syntax you proposed, but thank you for suggestion!

legal90:
@alexsomesan, @rporres, @meyskens Is anyone from maintainers interested in moving that forward (as well as `import` feature) ?

Issue #188, opened 2019-01-26
Istio Snapshot.3 fails to install


aaronmell:
When trying to install istio via the plugin, it fails. See https://discuss.istio.io/t/unable-to-install-istio-with-terraform-helm-plugin/507 for an overview of the issue. I originally thought that adding the configuration after installing istio also failed. But that seems to fail with both HELM and the pluging

This only seems to fail in AWS. Running locally on docker kubernetes, I have no issue.

### Terraform Version
Terraform v0.11.11
+ provider.aws v1.23.0
+ provider.helm v0.7.0
+ provider.kubernetes v1.5.0
+ provider.null v2.0.0

### Affected Resource(s)
- helm_release

### Terraform Configuration Files
```
resource "helm_release" "istio_bootstrap" {
  name       = "istio"
  namespace  = "${var.istio_namespace}"
  chart      = "./istio-1.1.0-snapshot.3/install/kubernetes/helm/istio"
  depends_on = ["null_resource.istio_download"]
  timeout    = 1200
}

resource "helm_release" "istio_configuration" {
  name       = "servicemesh"
  chart      = "chartmuseum/servicemesh"
  namespace  = "${var.istio_namespace}"
  #depends_on = ["helm_release.istio_bootstrap"]

  values = ["${var.istio_servicemesh_configuration}"]

  provisioner "local-exec" {
    command = "sleep 90s"
  }
}
```

aaronmell:
I was initially thinking that updating helm might fix this issue. Unfortunately It did not. @ @ #

aaronmell:
I identified my issue. Was missing an open port on 443 between the control plane and the worker nodes

PR #190, opened 2019-01-28
Updated dependencies to use Helm 2.12.3, Terraform 11.11 and kubernetes 1.12.0

aaronmell:
I thought this might fix issue #188 but it didn't seem to have any effect.


meyskens:
Going to close this because we have to use Go modules in Terraform 0.12

PR #193, opened 2019-01-30
helm_release: Suppress diff of "keyring" and "devel" attributes

### LINKED ISSUE #52: Keyring constantly changing
Hi,

Is there a way to force an attribute for all instances of helm_release{}? Every time we run terraform as a different user it wants to change keyring in every helm_release, which is kind of annoying.  It's a value we don't set, but it defaults to your homedirectory.  Failing that, if it stored it as `~/.gnupg` instead of the expanded path, it wouldn't need to change for different users.

legal90:
- Suppress diff of `devel` attribute if `version` is set. In this case `devel` attribute is not used at all and its value shouldn't affect the resource state.

- Suppress diff of "keyring" attribute regardless of anything else. This attribute is only used for verifying chart on the download phase, and it is just a local path. So its value in the terraform state doesn't really matter.

Fixes #52

### DIFF
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -63,7 +63,11 @@ func resourceRelease() *schema.Resource {
 			"devel": {
 				Type:        schema.TypeBool,
 				Optional:    true,
-				Description: "Use chart development versions, too. Equivalent to version '>0.0.0-0'. If version is set, this is ignored",
+				Description: "Use chart development versions, too. Equivalent to version '>0.0.0-0'. If `version` is set, this is ignored",
+				// Suppress changes of this attribute if `version` is set
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					return d.Get("version").(string) != ""
+				},
 			},
 			"values": {
 				Type:        schema.TypeList,
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -144,7 +144,11 @@ func resourceRelease() *schema.Resource {
 				Type:        schema.TypeString,
 				Optional:    true,
 				Default:     os.ExpandEnv("$HOME/.gnupg/pubring.gpg"),
-				Description: "Location of public keys used for verification.",
+				Description: "Location of public keys used for verification. Used only if `verify` is true",
+				// Suppress changes of this attribute as it is just a local path
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					return true
+				},
 			},
 			"timeout": {
 				Type:        schema.TypeInt,
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -145,9 +145,9 @@ func resourceRelease() *schema.Resource {
 				Optional:    true,
 				Default:     os.ExpandEnv("$HOME/.gnupg/pubring.gpg"),
 				Description: "Location of public keys used for verification. Used only if `verify` is true",
-				// Suppress changes of this attribute as it is just a local path
+				// Suppress changes of this attribute if `verify` is false
 				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
-					return true
+					return !d.Get("verify").(bool)
 				},
 			},
 			"timeout": {


helm/resource_release.go
```diff
@@ -140,7 +144,11 @@ func resourceRelease() *schema.Resource {
 				Type:        schema.TypeString,
 				Optional:    true,
 				Default:     os.ExpandEnv("$HOME/.gnupg/pubring.gpg"),
-				Description: "Location of public keys used for verification.",
+				Description: "Location of public keys used for verification. Used only if `verify` is true",
+				// Suppress changes of this attribute as it is just a local path
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
```
rporres, line 149: I know it's a weird functionality for Terraform as it is local, but I'd rather retain this functionality if someone has the need for it. So we either set it to a default value that does not change, e.g `""` or we do not suppress changes when `verify` is true

rporres (changes requested):
Thanks for your contribution, @legal90. Would it be possible to add tests for this too?

helm/resource_release.go
legal90: > So we either set it to a default value that does not change, e.g `""` or we do not suppress changes when `verify` is true

@rporres Thanks for your reply! Yes, I was also thinking about these variants. But then I just din't find any reason why someone would need to store the local path to the keyring in the state (which is commonly shared). Do you know any cases when it is needed?

As I see, when several people work on the same terraform state, they should be able to use custom paths to the keyring (via variables, for example) and that should not cause a resource update. Otherwise we run into #52.

FYI: Regardless of changes of this attribute, its value from the actual configuration is always passed to the Helm API when `verify` is true. So chart verification will always work with the keyring which is currently specified in Terraform config. Neither the old value nor the diff doesn't affect the resource behavior underneath.

Or maybe I miss something?

rporres: I must say that I've never worked with chart verification myself, all I know is from reading the [helm provenance documentation](https://github.com/helm/helm/blob/master/docs/provenance.md). Reading that I understand where the default value comes from. I also think that if an organisation wants to work with chart verification, it should place the keyring (containing the public key associated with the signed chart) in the same place for every computer that runs terraform. That's why I think we should not ignore the changes in that value in case `verify` is true but we can safely ignore in any other case to avoid running into #52

Does that make sense?

legal90: Not really, I'm sorry :)

> That's why I think we should not ignore the changes in that value in case `verify` is true

As I said in the previous comment, the saved value of `keyring` attribute does not affect the resource behavior even when `verify` is true. Having this attribute changed doesn't necessarily mean that the keyring is changed (and vice versa). But luckily, the resource reads the actual content of the keyring file every time when it fetches the chart. That's why I thought we can safely ignore changes of the local path in all cases.

But OK, if you believe we should keep tracking this attribute for the case when `verify` is true, then no problem - I'll add that quickly. But then we don't eliminate #52 completely.

legal90:
> Would it be possible to add tests for this too?

I guess, that should be a unit test for the `terraform.State` (or `terraform.InstanceState`?). I tried to create it, but didn't find a proper example for that :(

helm/resource_release.go
rporres: The `keyring` value is apparently used to verify the chart [locally](https://github.com/terraform-providers/terraform-provider-helm/blob/master/helm/resource_release.go#L721-L729) or [remotely](https://github.com/terraform-providers/terraform-provider-helm/blob/master/helm/resource_release.go#L770-L779). I must say I haven't tested myself that code and there are no tests for that, so I cannot say if it's working properly or if there's a bug there. I will take the action point of testing that and file an issue if it doesn't work.

In the meantime, we'd rather assume that the feature should not be broken so we should keep tracking this attribute for the case when `verify` is true. Thanks!

legal90: OK. 👍
I pushed the new commit to the branch. It changes the `DiffSuppressFunc` logic.

rporres:
The real test here would be to have a test for verify... Since it's out of the scope of your change, I will merge it as it is. Thanks for your contribution.

rporres (approved):

legal90:
@rporres Thank you!

Issue #199, opened 2019-02-04
[PROPOSAL] Switch to Go Modules


appilon:
As part of the preparation for Terraform v0.12, we would like to migrate all providers to use [Go Modules](https://github.com/golang/go/wiki/Modules). We plan to continue checking dependencies into vendor/ to remain compatible with existing tooling/CI for a period of time, however go modules will be used for management. Go Modules is the official solution for the go programming language, we understand some providers might not want this change yet, however we encourage providers to begin looking towards the switch as this is how we will be managing all Go projects in the future. Would maintainers please react with :+1: for support, or :-1: if you wish to have this provider omitted from the first wave of pull requests. If your provider is in support, we would ask that you avoid merging any pull requests that mutate the dependencies while the Go Modules PR is open (in fact a total codefreeze would be even more helpful), otherwise we will need to close that PR and re-run `go mod init`. Once merged, dependencies can be added or updated as follows:

```
$ GO111MODULE=on go get github.com/some/module@master
$ GO111MODULE=on go mod tidy
$ GO111MODULE=on go mod vendor
```

GO111MODULE=on might be unnecessary depending on your environment, this example will fetch a module @ master and record it in your project's go.mod and go.sum files. It's a good idea to tidy up afterward and then copy the dependencies into vendor/. To remove dependencies from your project, simply remove all usage from your codebase and run:

```
$ GO111MODULE=on go mody tidy
$ GO111MODULE=on go mod vendor
```

Thank you sincerely for all your time, contributions, and cooperation!

alexsomesan:
Hi Alex,

Yeah, I'm aware this one needs to switch to go modules as well.
I've had a quick go at it a while back and `go mod init` didn't import the glide dependencies correctly.
I'll have another go at it and take some time to fix the errors.

meyskens:
Closing as duplicate of #196

PR #200, opened 2019-02-04
Go modules

alexsomesan:
This change transitions the repository to using Go modules for dependency management.
This is a prerequisite for the migration to Terraform 0.12.x.


alexsomesan:
Fixes #199

meyskens (approved):
Surprised that this works given all the dep trouble!

alexsomesan:
Actually, me too :)
I've tried a while back and failed to get it working. Now I tried harder :)

appilon (approved):

meyskens:
@alexsomesan missed this in the giant diff but you forgot to remove the glide and glide lock files

alexsomesan:
Yeah, thanks for pointing that out. I do need to align the vendored K8S packages to the old versions too. I'll follow up with a PR soon.

PR #203, opened 2019-02-07
Implement Tiller resource

### LINKED ISSUE #134: Add tiller resource
Hi,

I think it would be great if you could manage the deployment of tiller across different namespaces with a dedicated resource.

ctrox:
This implements a new resource to manage a Tiller installation. It uses `k8s.io/helm/cmd/helm/installer` the same way the provider does right now and supports installing/upgrading/removing Tiller.

Fixes #134


Jeeppler:
Bitnami evaluate the security of Tiller running in Kubernetes. One of their findings was, that Tiller can be accessed from any pod inside the cluster [1].  The remediation is to limit Tiller to only listen to localhost. By limiting Tiller to listen to localhost, the Helm CLI can still reach Tiller, but pods inside the cluster are unable to reach Tiller. A user already created a working example using the Kubernetes Provider, where he limits access to localhost only [2].

I think it is important to mitigate the risk of in-cluster attacks and limit Tiller to only listen to localhost. In my opinion the Tiller resource should have at least an option to listen only to localhost. I personally would prefer if the option would be set to `true` by default. 

Here is an example with the listen to localhost option:

```hcl
resource "helm_tiller" "install" {
  only_listen_to_localhost = true
  namespace       = "tiller"
  service_account = "default"
  tiller_image    = "gcr.io/kubernetes-helm/tiller:v2.11.0"
}
```

The parameter `only_listen_to_localhost` is just a suggestion, feel free to change it.

[1]:  [Exploring the Security of Helm - In-cluster attacks](https://engineering.bitnami.com/articles/helm-security.html#in-cluster-attacks)
[2]: [terraform-tinfoil-tiller](https://github.com/lsst-sqre/terraform-tinfoil-tiller/blob/master/main.tf#L36)

ctrox:
I agree that we should have secure defaults but I'm not sure if this can be in the scope of this Terraform resource. The resource is using the function `installler.Install` of the helm package. [This function](https://github.com/helm/helm/blob/master/cmd/helm/installer/install.go#L199) does not allow to set any command options to start tiller with the `--listen=localhost:44134` flag. So we would have to patch the deployment after creating it which I don't really like. Do you have another idea on how to inject that listen command?

Of course another way to make tiller secure is by configuring TLS auth, which is something you can already do. It is just not really easy to enable that by default.

yorinasub17:
I believe you can change the startup command of the container using https://github.com/helm/helm/blob/v2.11.0/cmd/helm/installer/options.go#L98

This corresponds to https://github.com/helm/helm/blob/v2.11.0/cmd/helm/init.go#L151, which is what you can use to override the container command.

So:
```
// var opts installer.Options
opts.Values = []string{
  "spec.template.spec.containers[0].command={/tiller,--listen=localhost:44134}",
}
```
should do the trick.

Reference: Using `--override` to set the storage backend https://docs.helm.sh/using_helm/#storage-backends , which uses the same pathway to pass `--storage=secret` to the `tiller` command.

Jeeppler:
@ctrox if is difficult to set a parameter via command line to only listen to localhost, than we could simply work together with the Helm community and make sure the parameter can be set easily in one of the next releases.

However, let's see first if @yorinasub17 solution works.

ctrox:
Nice find @yorinasub17. I called the parameter just `listen_localhost` and defaulted it to `true`. It seems to work great so far. One issue with the resource that I have noticed while adding this is that when you change something, it will call `installer.Upgrade` which only applies changes to a few selected options (image, pull policy and serviceaccount). So for now I have added `ForceNew` to the `listen_localhost` parameter so it will uninstall/install tiller completely. What do you think about this? I guess if we go that way we would also need to add `ForceNew` to every parameter that is not one of the ones `installer.Upgrade` touches.

Edit: The acceptance test failed because of some minikube issue :man_shrugging:

ctrox:
I rebased on master and added `ForceNew` to all the params that a helm upgrade does not touch. Please let me know if anything else is missing for this to get merged.

Jeeppler:
@meyskens and @mcuadros can this pull request be merged?

rporres:
First of all, apologies for the late review and thanks a lot for your contribution.

The PR adds a tiller resource that is not connected to the helm release in any way. Tiller should be connected to the releases the same way it is done in `helm` command: it uses the `kube-system` tiller unless an alternative one is provided, so that would mean adding a `tiller_namespace` to the `helm_release` resource. Connection handling should be also changed as we may have multiple connections from different releases to various tillers.

It also doesn't address the issue of handling the lifecycle of the current tiller that's installed by default, which is a big problem as tiller cannot be updated from terraform. Current provider tiller installation should be deprecated now that we have a real tiller resource.

The additions proposed here are substantial but would make the provider benefit from a tiller handling that's similar to `helm` command. It may prove a fair amount of work to do, please let us know if you would prefer to have collaborators on this.

ctrox:
Thanks for having a look @rporres.

This certainly makes sense if we are going to deprecate the provider tiller installation. So the `GetHelmClient()` etc. from the provider would move to the release resource, and connect to whatever tiller you specify with `tiller_namespace`? But thinking about this some more, this would also mean we would need to configure authentication (if enabled) with each `helm_release` resource which I don't really like. I think authentication and connections should be handled by the provider as it is usually the case with terraform. You can still have multiple tiller installations if you create multiple provider instances with the use of aliases.

> It also doesn't address the issue of handling the lifecycle of the current tiller that's installed by default

I guess we could solve this by adding an import to the tiller resource? So to upgrade you would just set `install_tiller = false` in the provider (assuming we keep this for another release or two and just mark it deprecated) and import the existing tiller into the new resource. Afterwards you can upgrade tiller just by using the resource.

I would be happy to continue working on this but if someone wants to help to get this done faster that would of course be great as well.

sagikazarmark:
Installing tiller (the same way helm CLI would) is already possible using a deployment resource.

One can easily create a module like this: https://github.com/sagikazarmark/terraform-tiller

It allows setting custom options in the deployment, if necessary. TBH I find the deployment approach slightly better for that reason.

Jeeppler:
@sagikazarmark there is already a module for tiller: https://github.com/lsst-sqre/terraform-tinfoil-tiller

In addition, there are already two other modules in the Terraform registry:
- https://registry.terraform.io/modules/headcr4sh/tiller/kubernetes/
- https://registry.terraform.io/modules/Vidimensional/tiller/kubernetes/

sagikazarmark:
@Jeeppler sure, that's my point exactly: not sure what native tiller install support actually brings to the table.

rporres:
> But thinking about this some more, this would also mean we would need to configure authentication (if enabled) with each helm_release resource which I don't really like.

Good point. You would need to carry the credentials into the releases as they wouldn't be available through meta. That's not pretty, you're right

> I think authentication and connections should be handled by the provider as it is usually the case with terraform

My impression is that there's too much stuff in that provider. It should just delegate it to a library.

> You can still have multiple tiller installations if you create multiple provider instances with the use of aliases.

That's a very interesting workaround if we decided to keep the tiller where it is.

> I guess we could solve this by adding an import to the tiller resource? So to upgrade you would just set install_tiller = false in the provider (assuming we keep this for another release or two and just mark it deprecated) and import the existing tiller into the new resource. Afterwards you can upgrade tiller just by using the resource.

That's an interesting idea, but I don't get how we would solve the credentials situation in the case that we had multiple tiller resources. We would need multiple clients, anyway, isn't it?

> I would be happy to continue working on this but if someone wants to help to get this done faster that would of course be great as well.

Please, do.

rporres:
@sagikazarmark @Jeeppler The provider as it is allows for tiller to not be installed, but the expectation is for it to work as Helm does, managing the whole ecosystem, tiller included.

pankajagrawal16:
when will this be merged and which release is this targeted? 🤔

ctrox:
I am currently deploying tiller using a terraform `kubernetes_deployment` as described in some comment above. Seeing how helm is moving away from tiller in the next major version and the issues we are having here with multiple tiller instances, credentials handling etc. I would suggest to close this.

Jeeppler:
@ctrox I would prefer to keep this open to see where Helm is moving to and close it afterwards. I assume not everybody is able to move to the next major Helm version immediately.

aareet:
We're only accepting enhancements for Helm 3 from this point on.

PR #204, opened 2019-02-07
Update repository.html.markdown

ddi-wtag:


legal90:
In fact, `helm_repository` resource does not support import at all, so this section should be removed completely.

rporres:
Hi @ddi-wtag. Thanks for your contribution. @legal90 is right, we should state that `helm_repository` does not support import. Taking care of it as part of #209

PR #206, opened 2019-02-07
Add entries to .gitignore to roughly match the Google provider

rileykarson:
While playing with the provider locally, I noticed my IntelliJ `.idea` files were showing up in `git status`; use most of the [`.gitignore` we have in the Google provider](https://github.com/terraform-providers/terraform-provider-google/blob/master/.gitignore).


rporres:
Thanks!

PR #209, opened 2019-02-08
Updated documentation

rporres:
* Added missing attributes
* Reordered to make it easier to check attributes with the code
* Removed importing commands which do not work for the moment


legal90:
Maybe we should also make the example snippet more detailed? Like I tried to do here:
https://github.com/terraform-providers/terraform-provider-helm/pull/184/files#diff-7e12a07998cff3f121044250fda42c9eR19

rporres:
Sure, @legal90. I will make the example snippet more detailed

meyskens (approved):

PR #219, opened 2019-02-14
Migrate provider to Terraform 0.12 SDK

alexsomesan:
This change prepares the provider for compatibility with Terraform 0.12+

Notable changes include:
* Terraform provider SDK (via `github.com/hashicorp/terraform` module) vendored from close to tip of master.
* Vendor in Helm v2.11.0 and relevant Kubernetes components from release-1.1.
* Cleanup and reconcile various transient dependencies.

This is a major change and requires some burn-in testing. This PR aims to surface failing tests and address the issues until we reach a clean build.

Adititionaly, before merging, the final version of the Terraform SDK as released in the upcoming TF 0.12 should be vendored in. In the meantime we try to track Terraform master as closely as possible.


bflad (approved):
LGTM! 🚀

PR #227, opened 2019-02-28
Escape commas in set_string

### LINKED ISSUE #226: set_string parsing error
### Terraform Version

```
❯ terraform -v
Terraform v0.11.11
+ provider.cloudflare v1.11.0
+ provider.google v1.20.0
+ provider.google-beta v1.20.0
+ provider.helm v0.8.0
+ provider.kubernetes v1.5.0
+ provider.null v2.0.0
+ provider.random v2.0.0
+ provider.template v2.0.0
```

### Affected Resource(s)

- helm_release

```hcl
resource "helm_release" "nginx_ingress" {
  name      = "nginx-ingress"
  chart     = "stable/nginx-ingress"
  version   = "1.3.0"
  namespace = "nginx-ingress"

  set_string = {
    # name = "controller.config.proxy-real-ip-cidr"
    name  = "test"
    value = "${join(",", data.cloudflare_ip_ranges.cloudflare.cidr_blocks)}"
  }
}
```

### Panic Output

```
❯ terraform apply "tfplan"
Acquiring state lock. This may take a few moments...
Releasing state lock. This may take a few moments...
module.main_v2.helm_release.nginx_ingress: Modifying... (ID: nginx-ingress)
  set_string.#:                "0" => "1"
  set_string.2608104940.name:  "" => "test"
  set_string.2608104940.value: "" => "10.0.0.0/32,10.0.0.1/32,10.0.0.2/32"
Releasing state lock. This may take a few moments...

Error: Error applying plan:

1 error(s) occurred:

* module.main_v2.helm_release.nginx_ingress: 1 error(s) occurred:

* helm_release.nginx_ingress: failed parsing key "test" with value 10.0.0.0/32,10.0.0.1/32,10.0.0.2/32, key "1/32" has no value (canno
t end with ,)

Terraform does not automatically rollback in the face of errors.
Instead, your Terraform state file has been partially updated with
any resources that successfully completed. Please address the error
above and apply again to incrementally change your infrastructure.
```

### Expected Behavior

Should apply

### Actual Behavior

Get Error

### Steps to Reproduce

1. `terraform apply`

meyskens:
Fixes #226 

Regex advice welcome, what I wanted to use isn't available in Go

### DIFF
--- a/website/docs/release.html.markdown
+++ b/website/docs/release.html.markdown
@@ -72,7 +72,7 @@ The `set`, `set_sensitive` and `set_strings` blocks support:
 * `name` - (Required) full name of the variable to be set.
 * `value` - (Required) value of the variable to be set.
 
-Note: Unlike `--set` in Helm commas in `value` blocks are supported and will not be seen as a new key value set. 
+Note: Unlike `--set` in `helm` command, commas in `value` blocks are supported and will not be seen as a new key value set and don't need to be escaped as the provider will do it. This departure from the behaviour of the `helm` command is motivated by the fact that we cannot use a single `set` block for multiple keys in the provider
 
 ## Attributes Reference
 
--- a/website/docs/release.html.markdown
+++ b/website/docs/release.html.markdown
@@ -72,6 +72,7 @@ The `set`, `set_sensitive` and `set_strings` blocks support:
 * `name` - (Required) full name of the variable to be set.
 * `value` - (Required) value of the variable to be set.
 
+Note: Unlike `--set` in Helm commas in `value` blocks are supported and will not be seen as a new key value set. 
 
 ## Attributes Reference
 
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -578,6 +578,7 @@ func getValues(d *schema.ResourceData) ([]byte, error) {
 
 		name := set["name"].(string)
 		value := set["value"].(string)
+		value = nonEscapedCommaRegexp.ReplaceAllString(value, "$1\\,") // escape any non-escaped commas
 
 		if err := strvals.ParseInto(fmt.Sprintf("%s=%s", name, value), base); err != nil {
 			return nil, fmt.Errorf("failed parsing key %q with value %s, %s", name, value, err)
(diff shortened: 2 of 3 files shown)


rporres:
aren't we departing from `helm` command with this change? Shouldn't commas be escaped from the client?

meyskens:
Helm uses commas in `--set key1=val1,key2=val2`. 
Our code will never be able to 
```hcl
set {
  name = "key1,key2"
  value = "val1,val2"
}
```
As that will be `key1,key2=val1,val2`

rporres:
Sure, but there's an expectation of the provider to behave like `helm`...

rporres:
Anyway this may be a case in which it makes sense to depart from `helm` for the benefit of the user. In this case why not

* Add it to `set` and `set_sensitive`
* Add tests in which we have escaped commas (to avoid breaking backwards compatibility and/or a similar experience as `helm` would provide)
* Add a explaining note in docs

meyskens:
Added it everywhere with more tests. Feedback on documentation certainly welcome!

website/docs/release.html.markdown
```diff
@@ -72,6 +72,7 @@ The `set`, `set_sensitive` and `set_strings` blocks support:
 * `name` - (Required) full name of the variable to be set.
 * `value` - (Required) value of the variable to be set.
 
+Note: Unlike `--set` in Helm commas in `value` blocks are supported and will not be seen as a new key value set.
```
rporres:
```suggestion
Note: Unlike `--set` in `helm` command, commas in `value` blocks are supported and will not be seen as a new key value set and don't need to be escaped as the provider will do it. This departure from the behaviour of the `helm` command is motivated by the fact that we cannot use a single `set` block for multiple keys in the provider
```

rporres: What do you think, @meyskens?

mwconceicao:
@meyskens Hi! Did you mean to add this to `set` also? Because I think this broke lists definition on `set` attribute (syntax: `{a,b,c}`).

Should I open an issue to discuss this?

/cc @rporres

rporres:
sure, open an issue with a case @mwconceicao

Issue #68, opened 2018-04-24
No in-cluster authentication support?


jsleeio:
We deploy stuff with Buildkite agents that run in Kubernetes pods, therefore serviceaccount authetication is used.

I had a look through `terraform-provider-helm` code here and it looked like in-cluster auth is not supported.

If you don't already have this somewhere on a roadmap, I'll get to work on a PR. Thanks for creating and sharing this provider!

Related: should the Kube API endpoint default to `https://kubernetes` ?

mcuadros:
A PR is always welcome!

rporres:
This is still pending as #76 does not use https://godoc.org/k8s.io/client-go/rest#InClusterConfig to get the credentials k8s gives to pods

mcuadros:
Closing this issue since is making reference to a version based on Helm 2, if this is still valid to the master branch please reopen it. Thanks.

PR #231, opened 2019-03-07
Added load_config_file option to the provider k8s part

### LINKED ISSUE #230: Adding load_config_file argument
I want to be able to ignore the local kubernetes config file

### Terraform Version
```
Terraform v0.11.7
+ provider.helm v0.8.0
```

### Affected Resource(s)
- Helm provider

```hcl
provider "helm" {
  kubernetes {
    host = "${data.google_container_cluster.default.endpoint}"
    insecure = true

    username = "${data.google_container_cluster.default.master_auth.0.username}"
    password = "${data.google_container_cluster.default.master_auth.0.password}"
  }
}

```

### Expected Behavior
The provider should ignore the certificates in the kubernetes config file.

### Actual Behavior
```
Error: Error refreshing state: 1 error(s) occurred:

* provider.helm: failed to configure kubernetes config: specifying a root certificates file with the insecure flag is not allowed
```
PS:  These configs work fine if I remove the kubernetes config file `~/.kube/config`

## References
- https://www.terraform.io/docs/providers/kubernetes/index.html#load_config_file

meyskens:
Fixes #230

### DIFF
--- a/helm/provider.go
+++ b/helm/provider.go
@@ -215,6 +215,12 @@ func kubernetesResource() *schema.Resource {
 				Optional:    true,
 				Description: "Retrieve config from Kubernetes cluster.",
 			},
+			"load_config_file": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("KUBE_LOAD_CONFIG_FILE", true),
+				Description: "By default the local config (~/.kube/config) is loaded when you use this provider. This option at false disable this behaviour.",
+			},
 		},
 	}
 }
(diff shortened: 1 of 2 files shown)


rporres:
Shouldn’t this be feasible setting `in_cluster` to `false`? I mean, I like much more the `load_config_file` name, but I fail to see the difference between both or why we should keep both (when it would be useful, I mean)

We may want to deprecate `in_cluster` in favour of `load_config_file` so that we're in-sync with the k8s provider

rporres:
The problem here is that `in_cluster` is kinda broken. It should avoid loading `.kube/config` as it should call kubernets `InClusterConfig` function: https://godoc.org/k8s.io/client-go/rest#InClusterConfig

There can be two situations why you don't want to load  `.kube/config`

* You're running inside a pod and you want to authenticate using the service account k8s gives to pods
* You don't want it for your own personal reasons

Let's merge this as it is. I will update #68 to signal that it has still work to do

PR #238, opened 2019-03-13
Codify infrastructure for running acceptance tests

radeksimko:
This is not meant to replace the acceptance test setup in Travis, at least not at this point. We can look into a solution which satisfies all needs (e.g. security, transparency, convenience) later.

For now this enables us to run tests in TeamCity, where we test all other providers:

Screen Shot 2019-03-13 at 14 22 55

and also aligns this provider with others:
- https://github.com/terraform-providers/terraform-provider-panos/tree/master/panos/test-infra
- https://github.com/terraform-providers/terraform-provider-kubernetes/tree/master/kubernetes/test-infra

### DIFF
--- a/helm/test-infra/gke/kubeconfig-template.yaml
+++ b/helm/test-infra/gke/kubeconfig-template.yaml
@@ -1,0 +1,23 @@
+apiVersion: v1
+kind: Config
+preferences:
+  colors: true
+current-context: tf-helm-gcp-test
+contexts:
+- context:
+    cluster: ${cluster_name}
+    namespace: default
+    user: ${user_name}
+  name: tf-helm-gcp-test
+clusters:
+- cluster:
+    server: https://${endpoint}
+    certificate-authority-data: ${cluster_ca}
+  name: ${cluster_name}
+users:
+- name: ${user_name}
+  user:
+    password: ${user_password}
+    username: ${user_name}
+    client-certificate-data: ${client_cert}
+    client-key-data: ${client_cert_key}
--- a/helm/test-infra/install-helm-via-docker.sh
+++ b/helm/test-infra/install-helm-via-docker.sh
@@ -1,0 +1,62 @@
+#!/bin/bash
+set -e
+
+which docker > /dev/null
+if [ "$?" != "0" ]; then
+	echo "Error: Docker not found"
+	exit 1
+fi
+if [ "$KUBE_DIR" == "" ]; then
+	echo "Error: Please specify KUBE_DIR"
+	exit 1
+fi
+if [ "$HELM_VERSION" == "" ]; then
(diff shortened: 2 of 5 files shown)


helm/test-infra/install-helm-via-docker.sh
```diff
@@ -0,0 +1,58 @@
...
+	KUBECTL_VERSION=latest
+fi
+
+echo "HELM_HOME=${HELM_HOME}"
+
+docker version
+
+DOCKER_KUBECTL="docker run -i --rm \
+	-v ${KUBE_DIR}:/root/.kube \
+	lachlanevenson/k8s-kubectl:latest"
```
alexsomesan: I'm a bit uneasy about using this random container image from "the internetz" for kubectl.
May I suggest we use something more official, like Kubernetes' own images?
For example `k8s.gcr.io/hyperkube:v1.13.3` ([Docs](https://github.com/kubernetes/kubernetes/tree/master/cluster/images/hyperkube))

alexsomesan (commented):
Radek, thanks a lot for taking care of this.
All looks great, but I do have one spot issue with the source of the container images used.
Otherwise LGTM

helm/test-infra/install-helm-via-docker.sh
radeksimko: Good point.
I wasn't aware of official containers as I only searched through Docker Hub.

I will look into replacing these. 👍

radeksimko: Fixed.

helm/test-infra/gke/kubeconfig-template.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: v1
+kind: Config
+preferences:
+  colors: true
+current-context: tf-k8s-gcp-test
```
alexsomesan: Small nit: `tf-helm-gcp-test`

helm/test-infra/gke/kubeconfig-template.yaml
```diff
@@ -0,0 +1,23 @@
...
+contexts:
+- context:
+    cluster: ${cluster_name}
+    namespace: default
+    user: ${user_name}
+  name: tf-k8s-gcp-test
```
alexsomesan: tf-helm-gcp-test

helm/test-infra/gke/kubeconfig-template.yaml
radeksimko: Fixed 😉

helm/test-infra/gke/kubeconfig-template.yaml
radeksimko: Fixed 😉

alexsomesan (approved):
LGTM

PR #283, opened 2019-05-22
Fail on invalid chart

AndiDog:
Current situtation: if the chart is malformed, but the Helm release was already installed earlier, `terraform apply` will not display any error. I have a case where no `resource "helm_release" "foo" { version = ...` is given, so always the latest chart version should be taken. Terraform would then not hint that the chart is malformed, but silently "succeed", i.e. return an empty plan/apply result and the chart upgrade won't be attempted.

This patch fixes that and adds two tests (problems in `Chart.yaml` and `requirements.yaml`, respectively).

### DIFF
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -306,10 +306,10 @@ func resourceDiff(d *schema.ResourceDiff, meta interface{}) error {
 		return err
 	}
 
-	// Get Chart metadata, if we fail - we're done
+	// Get Chart metadata, if we fail - Terraform should fail
 	c, _, err := getChart(d, meta.(*Meta))
 	if err != nil {
-		return nil
+		return fmt.Errorf("failed to get chart metadata (malformed chart?): %v", err)
 	}
 
 	// Set desired version from the Chart metadata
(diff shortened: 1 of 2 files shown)


AndiDog:
What about the failed test?

```
--- FAIL: TestAccResourceRelease_repository (0.19s)
    testing.go:568: Step 0 error: errors during plan:
        
        Error: failed to get chart metadata (malformed chart?): Non-absolute URLs should be in form of repo_name/path_to_chart, got: coredns
```

I feel this might be a bug in `func (c *ChartDownloader) ResolveChartVersion(ref, version string) (*url.URL, getter.Getter, error)`. These are the resources used for the test:

```
func testAccHelmReleaseConfigRepository(resource, ns, name string) string {
	return fmt.Sprintf(`
		resource "helm_repository" "stable_repo" {
			name = "stable-repo"
		}

		resource "helm_release" %q {
			name       = %q
			namespace  = %q
			repository = "${helm_repository.stable_repo.metadata.0.name}"
			chart      = "coredns"
		}
	`, resource, name, ns)
}
```

and the function should consider the helm repo URL correctly, no?

AndiDog:
Any input why the test would fail?

PR #295, opened 2019-06-20
Wait for tiller if it's not ready

dstockton:
If tiller has failed for some reason; wait a reasonable amount of time to see if the pod/service will return to action.

### DIFF
--- a/helm/provider.go
+++ b/helm/provider.go
@@ -500,6 +500,13 @@ func (m *Meta) buildTunnel(d *schema.ResourceData) error {
 		return nil
 	}
 
+	// Wait a reasonable time for tiller, even if we didn't deploy it this run
+	o := &installer.Options{}
+	o.Namespace = d.Get("namespace").(string)
+	if err := m.waitForTiller(o); err != nil {
+		return err
+	}
+
 	var err error
 	m.Tunnel, err = portforwarder.New(m.Settings.TillerNamespace, m.K8sClient, m.K8sConfig)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


dstockton:
@alexsomesan @meyskens - sorry if you're not the correct people to contact. It'd be great if you'd be willing to review this XS PR to help with having the provider wait if tiller isn't available immediately
My use-case was tiller (not installed by terraform provider) hadn't started-up yet.

bmcustodio:
I was going to open a PR with this very same piece of code, since I am also facing issues with the provider thinking Tiller is ready when in fact it is not - and this seems to fix it. In my case, however, Tiller is installed by the provider, so there may be more than one issue here. Thoughts?

dstockton:
I haven’t checked but I think I’ve seen this before too - was thinking perhaps if there’s 2+ helm_release resources the first may install tiller, the 2nd skip it but tiller wasn’t ready. Either way, this PR resolves all my issues so far. :)

helm/provider.go
```diff
@@ -500,6 +500,13 @@ func (m *Meta) buildTunnel(d *schema.ResourceData) error {
 		return nil
 	}
 
+	// Wait a reasonable time for tiller, even if we didn't deploy it this run
+	o := &installer.Options{}
+	o.Namespace = d.Get("namespace").(string)
```
meyskens: Can we make this `m.Settings.TillerNamespace` to be consistent with the rest of the code in this function? 
```suggestion
	o.Namespace = m.Settings.TillerNamespace
```

dstockton: @meyskens I've updated with your suggestion. I'd used the same code as on line 437 for your reference. 
https://github.com/terraform-providers/terraform-provider-helm/blob/master/helm/provider.go#L437
Thanks for the review - hopefully you're able to approve. 😸

meyskens (approved):
Looking good! Thank you very much!

dstockton:
> Looking good! Thank you very much!

Thanks for the merge

Issue #265, opened 2019-04-30
helm_release "set" regression when setting a list


mwconceicao:
### Terraform Version
```
Terraform v0.11.13
+ provider.aws v2.8.0
+ provider.helm v0.9.1
+ provider.kubernetes v1.6.2
+ provider.random v2.1.1
```

### Affected Resource(s)
- helm_release

### Terraform Configuration Files
```hcl
provider "helm" {}

locals {
  mykeys = ["aaa", "bbb", "ccc"]
}

data "helm_repository" "stable" {
  name = "stable"
  url  = "https://kubernetes-charts.storage.googleapis.com"
}

resource "helm_release" "app" {
  repository = "${data.helm_repository.stable.name}"
  name       = "myapp"
  chart      = "stable/unbound"

  reuse_values = true

  set {
    name  = "myinfo.keys"
    value = "{${join(",", local.mykeys)}}"
  }
}
```

### Panic Output
_No panic._

### Expected Behavior
1. Apply runs successfully;
2. Release has `myinfo.keys` set as a list:
```
$ helm get values myapp
myinfo:
  keys:
  - aaa
  - bbb
  - ccc
```
3. My chart iterates successfully on `myinfo.keys` to add some environment variables on my app.

### Actual Behavior
Commas get escaped and treated as **one big item** making my chart to try to add an invalid config key on Kubernetes, because of the commas, causing the apply to fail.
```
$ helm get values myapp
myinfo:
  keys:
  - aaa,bbb,ccc
```

Apply error: https://gist.github.com/mwconceicao/1f4c5275893da5b82b8fecfd292102f6

### Steps to Reproduce
1. `terraform apply`

### Important Factoids

Problem started after PR #227.

### References

- Issue #226 
- PR #227 

---

Basically the problem started after PR #227 which adds the comma escaping code to [set][], [set_sensitive][] and [set_string][].

Not sure if this was intended to be added in those 3 sets. I think this comma escaping makes sense for `set_string` and perhaps to `set_sensitive` as well, but I think we should keep `set` without comma escaping so we don't brake list definitions (syntax `{aaa,bbb,ccc}`) which is something important to support.

I'm willing to open a PR for this.

What do you folks think?

Thanks!

[set]: https://github.com/terraform-providers/terraform-provider-helm/pull/227/files#diff-70e42ed4d5009a76bddc6fb6c16b913bR581
[...]

PR #302, opened 2019-07-02
Add set_list argument to helm_release

eduardocardoso:
Since PR #227 was merged it became impossible to send a list of values using the same syntax as with the helm cli (`key={val1,val2,val3}`). 
This adds a `set_list` argument that receives a list as its value to address issue #265.


rporres:
Thanks for the fix, but we'd better close this since we rolled back #227 in https://github.com/terraform-providers/terraform-provider-helm/pull/310

PR #304, opened 2019-07-03
Use GetOkExists instead of GetOk for boolean attributes

osdemah:
Workaround for #303

Issue #195, opened 2019-01-31
Provider errors during plan when k8s token is provided claiming username was also provided


iliasbertsimas:
### Terraform Version
v0.11.1

### Affected Resource(s)
- helm provider

### Terraform Configuration Files
```hcl
provider "helm" {

    version = "~> 0.7.0"

    //debug = true

    install_tiller = false
    service_account = "tiller"
    namespace = "${local.namespace}"

    kubernetes {
      host                              = "${var.eks_endpoint}"
      cluster_ca_certificate  = "${base64decode(var.eks_ca_data)}"
      token                            = "${data.external.aws-iam-authenticator.result.token}"
    }
}
```

### Debug Output
```
Error: Error refreshing state: 1 error(s) occurred:

* provider.helm: failed to configure kubernetes config: username/password or bearer token may be set, but not both
```
### Panic Output
none

### Expected Behavior
It should have used the provided token and completed the plan with success.

### Actual Behavior
It acts like username and password along with token were provided.

### Steps to Reproduce
1. `terraform plan`

### Important Factoids
Running on AWS EKS.

btai24:
I'm running into this as well.

gordonbondon:
Same issue with provider version `0.10.0`

gordonbondon:
I've submitted a PR with a partial fix, but this also requires #304 to be merged so `load_config_file = false` is actually correctly detected.

rporres:
Closing this. Feel free to reopen if the #307 doesn't fix it.

vliubko:
terraform version **0.12.12**
helm provider version **0.10.4**

The problem appeared, earlier with `0.10.2` version of helm-provider it doesn't exist.

P.S. The issue appears while working with Atlantis, on my local MacOS everything fine with **0.10.4**

vliubko:
@rporres should I open a new ticket or you will re-open this one?

gordonbondon:
@vliubko what's your provider config? are you getting the same `username/password or bearer token may be set` error? 

We've been running with two of these fixes for some time both locally and in Atlantis, and it worked fine for us.

vliubko:
@gordonbondon 
1) Yes, error is the same. `failed to configure kubernetes config: username/password or bearer token may be set, but not both`
2) config:
```
provider "helm" {
  version         = "0.10.4"
  tiller_image    = "gcr.io/kubernetes-helm/tiller:v2.14.0"
  service_account = var.service_account
  kubernetes {
[...]

PR #307, opened 2019-07-25
Fix loading kubeconfig when disabled

gordonbondon:
Fix for #195 

Also requires #304 to work properly.

Because of `clientcmd.NewDefaultClientConfigLoadingRules()` `c.ClientConfig()` was always trying to load kubeconfig from default locations, even if `load_config_file` was set to false. 

I moved all of this under if condition (similar to how its done in kubernetes provider https://github.com/terraform-providers/terraform-provider-kubernetes/blob/master/kubernetes/provider.go#L175) and now it works as expected.

### DIFF
--- a/helm/provider.go
+++ b/helm/provider.go
@@ -291,17 +291,9 @@ func (m *Meta) buildSettings(d *schema.ResourceData) {
 func (m *Meta) buildK8sClient(d *schema.ResourceData, terraformVersion string) error {
 	_, hasStatic := d.GetOk("kubernetes")
 
-	c, err := getK8sConfig(d)
+	cfg, err := getK8sConfig(d)
 	if err != nil {
 		debug("could not get Kubernetes config: %s", err)
-		if !hasStatic {
-			return err
-		}
-	}
-
-	cfg, err := c.ClientConfig()
-	if err != nil {
-		debug("could not get Kubernetes client config: %s", err)
 		if !hasStatic {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


pierresteiner:
Any chance to have this one merged soon? The underlying issue is really annoying

mzupan:
Can this be looked at? It's almost impossible to use terraform locally to test things like plans when this is broken and still reading the local kubectl config when it's referencing another cluster

gordonbondon:
Rebased and fixed the conflict.

landro:
A workaround that seems to work for me
 
```    config_path            = "nowhere"```

jmorgan415:
FYI, this is not fixed (for me at least).  I'm using helm provider 0.10.4 and Terraform 0.12.9.  I had to add the standalone kubernetes provider in addition to the kubernetes block in the helm provider to get any of my kubernetes resources to work, but helm still tries to load a config file if I don't include the standalone kubernetes provider.  In the past, I just used the helm provider with the nested kubernetes block for all k8s/helm resources, so this doesn't seem to be working anymore.

gordonbondon:
@jmorgan415 kubernetes provider and helm provider work separately and they cant affect each other in any way. 

Removing `kubernetes` block from helm provider configuration will cause it to use defaults, and loading config is the default behaviour, so this is expected.

jmorgan415:
@gordonbondon Thanks for the quick response.  I edited my comment above to reflect that I used to be able to use the helm provider with the nested kubernetes block for all k8s/helm resources.

gordonbondon:
@jmorgan415 can you post your config here?

PR #324, opened 2019-08-26
Modernise docs

rporres:
* Use 0.12 syntax
* Use two space indentation as suggested by
  https://www.terraform.io/docs/configuration/style.html

### DIFF
--- a/website/docs/index.html.markdown
+++ b/website/docs/index.html.markdown
@@ -22,18 +22,18 @@ The Helm provider is used to deploy software packages in Kubernetes. The provide
 
 ```hcl
 resource "helm_release" "mydatabase" {
-    name      = "mydatabase"
-    chart     = "stable/mariadb"
+  name  = "mydatabase"
+  chart = "stable/mariadb"
 
-    set {
-        name  = "mariadbUser"
-        value = "foo"
-    }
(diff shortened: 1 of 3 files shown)


website/docs/index.html.markdown
```diff
@@ -64,15 +64,15 @@ The other way is **statically** define all the credentials:
...
-        client_certificate     = "${file("~/.kube/client-cert.pem")}"
-        client_key             = "${file("~/.kube/client-key.pem")}"
-        cluster_ca_certificate = "${file("~/.kube/cluster-ca-cert.pem")}"
-    }
+  kubernetes {
+    host     = "https://104.196.242.174"
+    username = "ClusterMaster"
+    password = "MindTheGap"
+
+    client_certificate     = "${file("~/.kube/client-cert.pem")}"
```
meyskens: If we're making it 0.12 we should not use `"${}"`

meyskens (changes requested):
You forgot one change :wink:

website/docs/index.html.markdown
rporres: fixed in 1e69c0cbe

meyskens (approved):

PR #349, opened 2019-09-30
Use Kubernetes v1.15 for testing

meyskens:
There are some issues with the acceptance testing however we're behind a few versions of Kubernetes but use the latest minikube.
This also fixes some recent changes made in minikube that change how `update-context` works.


alexsomesan (approved):
LGTM

PR #350, opened 2019-09-30
Fix release shield

meyskens:
Since the change to the official Hashicorp release system we do not have github "releases" but we still have tags so the shield on the readme was incorrectly saying v0.6


invidian (approved):
LGTM

rporres (approved):

PR #359, opened 2019-10-20
Set up new Kubernetes port-forward for each Helm operation

shaneramey:
Sets up a new connection for each Helm release + operation pair, to work around a [known Kubernetes issue](https://github.com/kubernetes/kubernetes/issues/78446) where a large number of connections going through a single Kubernetes port-forward connection results in the connection being broken.

I was getting lots of "Context deadline exceeded" errors in EKS, which this PR fixes for me. 

The Helm CLI `helm install`, `helm get`, `helm update`, and `helm delete` commands worked fine, because these commands create a single port-forward for each operation/release. This PR replicates that behavior.

### DIFF
--- a/helm/provider.go
+++ b/helm/provider.go
@@ -519,9 +519,6 @@ func (m *Meta) waitForTiller(o *installer.Options) error {
 }
 
 func (m *Meta) buildTunnel(d *schema.ResourceData) error {
-	if m.Settings.TillerHost != "" {
-		return nil
-	}
 
 	// Wait a reasonable time for tiller, even if we didn't deploy it this run
 	o := &installer.Options{}
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -321,6 +321,8 @@ func resourceReleaseCreate(d *schema.ResourceData, meta interface{}) error {
 	if err != nil {
 		return err
 	}
+	defer m.Tunnel.Close()
+
 	name := d.Get("name").(string)
 
 	if err = prepareTillerForNewRelease(d, c, name); err != nil {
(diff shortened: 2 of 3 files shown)


goblain:
I think this was there for a reason. Just removing the logic to skip this if already initialized might not be the best idea. For starters, port forwarder is not closed anywhere so new tunnels will stockpile through the run. I'm all for this, but we should close pf after it was used.

Another approach would be to keep this but add a verification if tunnel is ok and (re)create only if required.

meyskens:
I'm with @goblain on this one

jakesylvestre:
can we get this merged? This would be extremely helpful

shaneramey:
@goblain  Great point about too many unclosed connections building up. I've updated the PR to close connections after they are used.

jakesylvestre (approved):

mgar:
Any chance that we can get this merged and included in the new release?

iainlbc:
This would be very helpful, we are running in to the same issue on EKS

paaloeye:
@mcuadros @meyskens @alexsomesan @rporres May we have your attention here?

pkaramol:
Is there an expected date for this to be merged?

helm/provider.go
```diff
@@ -519,9 +519,6 @@ func (m *Meta) waitForTiller(o *installer.Options) error {
 }
 
 func (m *Meta) buildTunnel(d *schema.ResourceData) error {
-	if m.Settings.TillerHost != "" {
```
jaygridley, line 522: This cannot be just removed as it also handles the case when the host is passed into provider configuration. I would suggest creating logic for building the tunnel only when it should be used, meaning when host is not provided from the configuration.

aaronmell:
Can you update this PR to target the V2 branch?

mcuadros:
As described on the README.md, the master is based on Helm 3 for a few months, and we are not accepting any enhancement for branches supporting Helm 2. If this is relevant to Helm 3, and anyone is interested in continuing this work, please open a new pull request.

PR #364, opened 2019-10-25
Remove the install instructions

meyskens:
Since we're official (for a while now) we do not need the manual instructions any longer.
Users should just refer to the official Terraform documentation for installing providers.


alexsomesan (approved):
LGTM!
Thanks a lot, @meyskens

PR #361, opened 2019-10-22
Upgrade Helm to v2.15.1 and Kubernetes to v1.15.5

JulienBreux:
Hello world 👋

I think that's requested soon by the community.

Here, two upgrades :
* [Helm v2.15.1](https://github.com/helm/helm/releases/tag/v2.15.1)
* [Kubernetes 1.15.5](https://github.com/kubernetes/kubernetes/releases/tag/v1.15.5)

---

## Note

With the new version of Kubernetes, I must use `replace` 😒
If you've a better way, I'll take it with pleasur 🤗 

Sorry for big `vendor/` changes apply by `go mod vendor`.

---

## Info

Others PR to adapt Helm v3 and Kubernetes v1.16.x later. 😉

PR #365, opened 2019-10-28
update tiller image

rporres:
This was missing from #361


meyskens (approved):

k0ffee:
The updated tiller image breaks post-install jobs for me.

That is tiller 2.14.3 works, tiller 2.15.1 doesn't.

See the issue that was wrongly attributed to Istio SDS first:
https://github.com/istio/istio/issues/18596

Issue #355, opened 2019-10-14
context deadline exceeded on cnnection cut due to timeout


goblain:
### Terraform Version
v0.12.9

### Affected Resource(s)
- helm_resource
- helm_repository

### Terraform Configuration Files
```hcl
data "helm_repository" "stable" {
    name = "stable"
    url  = "https://kubernetes-charts.storage.googleapis.com"
}

resource "helm_release" "test01" {
  name       = "redis-release01"
  repository = "${data.helm_repository.stable.metadata.0.name}"
  chart      = "redis"
  set {
    name = "cluster.enabled"
    value = "false"
  }
}

resource "null_resource" "delay" {
  depends_on = ["helm_release.test01"]
  provisioner "local-exec" {
    command = "sleep 300s"
  }
}

resource "helm_release" "test02" {
  depends_on = ["null_resource.delay"]
  name       = "redis-release02"
  repository = "${data.helm_repository.stable.metadata.0.name}"
  chart      = "redis"
}

```
### Expected Behavior
both helm_releases are applied

### Actual Behavior
second helm_release instance is not applied due to 
```
Error: context deadline exceeded
```
which is a result of kubernetes portforwarder tunnel dropped from kube api on  our AKS clusters (~4min timeout on connections via Azure LoadBalancer). 

### Steps to Reproduce
1. `terraform apply`

### Important Factoids
Running on AKS

### Solutions
As portforwarder is initiated once when provider initializes, if something (like LB inactivity timeout) causes tunnel TCP connection to drop, this error is consistently thrown. The simplest solution that works for me is to implement a "keepalive" GetVersion requests that are sent every N seconds to make sure the tunnel stays alive. Retry/recovery from a failure would probably be preferred but a bit more complex to implement. Quickfix I use for my env is in https://github.com/tradeix/terraform-provider-helm/commit/f4348483ecf5ed7224abd4f00febb12b912f805e , if you're willing to merge this kind of workaround I can open a PR for this.

shaneramey:
I created https://github.com/terraform-providers/terraform-provider-helm/pull/359 which fixed this issue for me.

mcuadros:
Closing this issue since is making reference to a version based on Helm 2, if this is still valid to the master branch please reopen it. Thanks.

Issue #308, opened 2019-07-26
EOF Error causes incorrect TF state


aaronmell:
We are seeing random EOF errors when applying helm_releases. We are attempting to update the image tag. When the EOF issue happens, the image tag is not updated in helm, and the next time we run terraform, it doesn't retry to update the image tag.

`
  resource "helm_release" "example_release" {
  name       = "example_release_name"
  repository = "example_release_name_repo"
  chart      = "example_release_name_chart"
  namespace  = "example_release_name_namespace"
  depends_on = ["null_resource.some_null_resource"]
  version    = "example_release_name.version"
  values = [
  "yaml for chart loaded from file",
]

 set {
    name  = "containerDefinition.image.repository"
    value = "${var.container_image_repository}"
  }

  set {
    name  = "containerDefinition.image.tag"
    value = "${var.container_image_tag}"
  }
`

We have some additional sets in a typical config, but for this example they are not important.

Now when I look at the TF state file after the error occurs I see the following
"metadata.0.values" tag: 1.2.69\n   has the image tag that is currently deployed and matches helm

However the set has the value that we were attempting to change to, rather than the value that exists in helm. 
"set.4006280493.value": "1.2.70", 

This is what terraform does
If the Update callback returns with or without an error, the full state is saved. If the ID becomes blank, the resource is destroyed (even within an update, though this shouldn't happen except in error scenarios).

I'm guessing because we pass in 1.2.70 the next time we run, it sees that the value was already set, and doesn't do anything.  When we do a ReleaseRead operation we only update the version, namespace, status and metadata. Since the status is never set to failed, and the version of the chart didn't change, no update would be required.

I can think of a few solutions on how to fix it, but I'm not sure of the ramifications of making such a change.

1. If Helm fails for some reason during the deployment, and the status did not change of the release, set the status value to something like "Unknown" so it forces a redeploy the next time.

2. call d.set and update all of the values in d.set during a read operation. This feels like the more correct option, but its also a big change, and I am not sure of the reasoning behind not updating the set during a read. 

### Terraform Version
Terraform v0.11.13

### Affected Resource(s)
- helm_release

aaronmell:
I'm thinking about adding a retry to the provider, so if the EOF issue occurs, or the Context Deadline issue occurs, both which seem to be random failures, the provider can simply retry the operation. Any thoughts?

aaronmell:
[...]

PR #374, opened 2019-11-07
set connection_timeout parameter

mgar:
As a solution for both #355 and #308, I added a new parameter to set the timeout in seconds to the Helm client.

This PR could be closed in favour of #359 if you think that closing the tunnel each time is the preferred solution

### DIFF
--- a/helm/provider.go
+++ b/helm/provider.go
@@ -78,7 +78,7 @@ func Provider() terraform.ResourceProvider {
 				Type:        schema.TypeInt,
 				Optional:    true,
 				Default:     60,
-				Description: "Number of seconds Helm will wait to establish a connection to tiller.",
+				Description: "Number of seconds Helm will wait before timing out a connection to tiller.",
 			},
 			"service_account": {
 				Type:        schema.TypeString,
--- a/helm/provider.go
+++ b/helm/provider.go
@@ -77,7 +77,7 @@ func Provider() terraform.ResourceProvider {
 			"connection_timeout": {
 				Type:        schema.TypeInt,
 				Optional:    true,
-				Default:     5,
+				Default:     60,
 				Description: "Number of seconds Helm will wait to establish a connection to tiller.",
 			},
 			"service_account": {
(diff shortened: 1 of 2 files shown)


macytd (approved):

jakesylvestre:
Tbh, this is a good stop gap either way.

helm/provider.go
```diff
@@ -74,6 +74,12 @@ func Provider() terraform.ResourceProvider {
 				Default:     "gcr.io/kubernetes-helm/tiller:v2.15.1",
 				Description: "Tiller image to install.",
 			},
+			"connection_timeout": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Default:     60,
+				Description: "Number of seconds Helm will wait to establish a connection to tiller.",
```
jakesylvestre: I think "Number of seconds Helm will wait before timing out a connection to tiller." is more accurate here. This makes it sound like a wait

website/docs/index.html.markdown
```diff
@@ -89,6 +89,7 @@ The following arguments are supported:
 * `init_helm_home` - (Optional) Initialize Helm home directory configured by the `home` attribute if it is not already initialized, defaults to true.
 * `install_tiller` - (Optional) Install Tiller if it is not already installed. Defaults to `true`.
 * `tiller_image` - (Optional) Tiller image to install. Defaults to `gcr.io/kubernetes-helm/tiller:v2.15.1`.
+* `connection_timeout` - (Optional) Number of seconds Helm will wait to establish a connection to tiller. Defaults to `5`.
```
jakesylvestre: This should be 60. It's inconsistent with the actual default [here](https://github.com/terraform-providers/terraform-provider-helm/pull/374/files#diff-9f0ddb4b472596921b8dc5a236a017b2R81)

jakesylvestre (changes requested):

website/docs/index.html.markdown
mgar: good catch!

meyskens:
Thanks for the PR! Even if we go for what was done in #359 I think this is a must to have configurable.

meyskens (approved):

grugnog:
@meyskens any chance we could tag a new release including this fix?

PR #375, opened 2019-11-07
Add flag disable_crd_hooks which allows use of the same --no-crd-…

philliprower:
Add flag disable_crd_hooks which allows use of the same --no-crd-hook offered via the command line helm.


meyskens:
@philliprower Thanks for the PR! Could you also add this to the documentation?

philliprower:
Sure.

meyskens (approved):

meyskens:
Thanks!

Issue #299, opened 2019-06-27
Support Helm 3.0


alexkreidler:
Helm 3.0 is in alpha and is heading towards being released. It would be great to start working on support for 3.0 so we could use it as soon as it comes out.

My main concern is it seems that this terraform plugin directly uses Tiller through RPCs. However, Helm 3.0 is eliminating tiller in favor of a client-side system that then directly uses the Kubernetes API. 

It seems like a significant refactor to move from the lower-level code that currently makes up this plugin to the types of simpler API calls to the helm library, which is why it would be good to start soon.

Thanks for a great plugin!

aaronmell:
3.0 is still in Alpha, Once the first beta drops, it should be worked on. I think working on it now might be painful.

enummela:
Just want to point out the Helm 3.0 beta was released today. Very much appreciate the efforts keeping this provider up to date. It will be exciting to follow the progress. Thanks!

https://github.com/helm/helm/releases/tag/v3.0.0-beta.1

aaronmell:
Anyone taking a stab at this? I might spend some time this weekend working on it.

aaronmell:
@rporres Asking you, since you are the most active contributor to the project. How should we handle the 3.0 work? I think we have should create a long running branch with the v3 work in it until it reaches GA. Should we have new providers/resources for v3, and rename the old resources to v2?

aaronmell:
I opened https://github.com/helm/helm/issues/6333 and submitted a PR https://github.com/helm/helm/pull/6341 to improve the configuration building in helm v3

aaronmell:
Thinking about it some more, there is no way to really support both helm 2 and helm 3 in a single version of the provider. I think bumping the major version is the correct answer, and maintaining a v2 branch for some period of time.

rporres:
@aaronmell I will find time to discuss with @alexsomesan how to organise this and come back to you

alexsomesan:
Now that Helm 3.0 is tagged as Beta, it meets the theoretical criteria to be considered for support in the provider. We should evaluate whether that comes with any API stability guarantees.

alexsomesan:
I agree with that we can't support both versions in the same provider version. We would keep maintaining v2 for as long as possible with Helm 2.0 support only and bump major for Helm 3.0 support.

aaronmell:
Still waiting on My PR to be accepted. Also still waiting on Go Module support to be merged into the helm 3.0 branch.

[...]

PR #378, opened 2019-11-23
Updating plugin to use helm version 3

aaronmell:
I'm opening the PR to solicit feedback from the community. Its not ready yet, but its close. I'll annotate some lines that I know need a working solution for. 

https://github.com/terraform-providers/terraform-provider-helm/issues/299


helm/provider_test.go
```diff
@@ -24,82 +26,163 @@ const (
...
-			"Acceptance tests skipped unless env '%s' set", resource.TestEnvVar))
-		return
+	if namespace != "" {
+		createNamespace(t, namespace)
 	}
 
-	helmHome, err := ioutil.TempDir("", "terraform-acc-test-helm-")
-	if err != nil {
-		t.Fatalf("Failed to create new temporary directory for use as helm home: %s", err)
+	os.Setenv("HELM_REPOSITORY_CONFIG", "/Users/amell/Library/Preferences/helm/repositories.yaml")
```
aaronmell: I could use some advice here. Not sure how these should be set.

helm/data_repository.go
```diff
@@ -75,83 +74,63 @@ func dataRepository() *schema.Resource {
...
+	} else {
+		file = repo.NewFile()
 	}
 
-	for _, r := range f.Repositories {
-		if r.Name == name {
-			return r, nil
+	entry := file.Get(name)
+
+	// Not sure I agree with the logic here. Should a data source really update an underlying resource every time its called?
```
aaronmell, line 97: Not sure about this behavior, and if it completely makes sense. Would like feed back.

helm/resource_release.go
```diff
@@ -45,7 +52,33 @@ func resourceRelease() *schema.Resource {
 			"repository": {
 				Type:        schema.TypeString,
 				Optional:    true,
-				Description: "Repository where to locate the requested chart. If is an URL the chart is installed without installing the repository.",
+				Description: "Repository where to locate the requested chart. If is a URL the chart is installed without installing the repository.",
+			},
+			"repository_key_file": {
```
aaronmell, line 47: Should the repository settings be moved into their own block of settings like kubernetes settings are in the provider?

aaronmell:
Seems to be an issue with pods not going into the correct namespace. I have a bunch of pods, and some of them ended up in the default namespace, even though I don't use the default namespace.

aaronmell:
> Seems to be an issue with pods not going into the correct namespace. I have a bunch of pods, and some of them ended up in the default namespace, even though I don't use the default namespace.

I know what the issue is here, working on a fix for it.

jakesylvestre:
I'm testing this out now, I'll let everyone know how it goes

aaronmell:
Theres is still a namespace bug that I haven't been able to squash yet.

relu:
> Theres is still a namespace bug that I haven't been able to squash yet.

Ha! Great timing, cause I've just sent over a PR for you: https://github.com/aaronmell/terraform-provider-helm/pull/3 :)

jakesylvestre:
@relu lol that's fantastic

and @aaronmell using a test cluster :)

aaronmell:
I've removed the WIP from the title. I think this close to being ready to merge in.

robinkb:
I actually may have found a bug in this PR.

I provision a Kubernetes cluster using AKS on Azure, and then use the credentials to intialize the Helm provider. This works fine using version 0.10.4 of this provider.

```hcl
provider "helm" {
  kubernetes {
    host                   = azurerm_kubernetes_cluster.kubernetes.kube_admin_config[0].host
    client_certificate     = base64decode(azurerm_kubernetes_cluster.kubernetes.kube_admin_config[0].client_certificate)
    client_key             = base64decode(azurerm_kubernetes_cluster.kubernetes.kube_admin_config[0].client_key)
    cluster_ca_certificate = base64decode(azurerm_kubernetes_cluster.kubernetes.kube_admin_config[0].cluster_ca_certificate)

    # We do not want to risk running Helm against the wrong cluster.
    load_config_file = false
  }
}
```

However, with the version built from your PR, I get this error:

```
Error: invalid configuration: [client-cert-data and client-cert are both specified for clusterAdmin_kubernetes_kubernetes. client-cert-data will override., client-key-data and client-key are both specified for clusterAdmin_kubernetes_kubernetes; client-key-data will override, certificate-authority-data and certificate-authority are both specified for kubernetes. certificate-authority-data will override.]
```

The errors are [surfaced from the Kubernetes client](https://github.com/kubernetes/client-go/blob/master/tools/clientcmd/validation.go#L212-L219).

It seems that the Helm provider is reading both the arguments that I passed in the `kubernetes` block, as well as my `~/.kube/config` file. The workaround shown below is able to initialize the Helm provider, so it seems to confirm my suspicion.

```hcl
// Generate a random string for naming the tempfile.
resource "random_string" "kubeconfig_tmpfile_name" {
  length  = 10
  special = false
}

// Write the raw kubeconfig for the provisioned Kubernetes cluster to the tempfile.
resource "local_file" "kubeconfig" {
  content  = azurerm_kubernetes_cluster.kubernetes.kube_admin_config_raw
  filename = "/tmp/tf_kubeconfig_${random_string.kubeconfig_tmpfile_name.result}"
}

provider "helm" {
  kubernetes {
    // Use the kubeconfig in the tempfile as configuration.
    config_path = local_file.kubeconfig.filename
  }
}
```

Besides this annoying bug, it seems to work quite well! :smile:

relu:
@robinkb please see the discussion started on the issue here: https://github.com/terraform-providers/terraform-provider-helm/issues/299#issuecomment-56015303 it is most certainly the same problem.
My last comment there contains a reference to a branch where I've implemented a sort of a hack to prevent the kuberentes client from loading the local kubeconfig.

robinkb:
@relu Ah, yup. That seems to be the same problem.

aaronmell:
I'm running into an issue, where is a release fails to deploy initially, the next run has the following error

* helm_release.bar: "bar" has no deployed releases

Helm ls shows the release
bar                    	namespace 	1       	2019-12-08 09:29:31.802095 -0600 CST	failed  	chart	1.0

my provider is 

`provider "helm" {
  version = "= 3.0.0"
  debug = true
  //insecure = true
}`

the resource I am using looks like this

`
resource "helm_release" "bar" {
  name       = "${local.microservice_name}"
  repository = "${var.helm_repository_name}"
  chart      = "local-repo/bar"
  namespace  = "${var.namespace}"
  version    = "${var.microservice_chart_version}"
  depends_on = ["null_resource.bar"]

  values = [
    "${file("../../bar/helm-config/bar.yaml")}"
  ]

//A bunch of different set values. 
  set {
    name  = "mode"
    value = "${var.node_env_variables["is_release"] ? "production" : "development" }"
  }
`

Pretty sure the issue here, is that we store a failed release in the state. So the next time we try to apply a failed release helm doesn't like it and aborts. The function in helm that checks before upgrading expects the release being upgraded to be a deployed state.

aaronmell:
@relu It looks like this change is causing the error I am seeing.
https://github.com/terraform-providers/terraform-provider-helm/pull/4
What was the issue with not storing failed state?

I added some information to this issue in the helm project, as it looks like a change in behavior from v2

https://github.com/helm/helm/issues/7160

aaronmell:
~~Looks like I removed the reuse option, ~~that needs to be readded~~, its now the replace flag. ~~

Edit: Merged in a change for this.

relu:
> @relu It looks like this change is causing the error I am seeing.
> #4
> What was the issue with not storing failed state?
> 
> I added some information to this issue in the helm project, as it looks like a change in behavior from v2
> 
> [helm/helm#7160](https://github.com/helm/helm/issues/7160)

The issue was that helm would create the "failed" release while it not being stored in the terraform state would cause re-executing apply to fail due to name collision since the provider would not recognize the old failed release and would require manual intervention via helm cli.

relu:
I see the tests are failing now, will see if I can fix them. I assume the cause is the updated helm package.

aaronmell:
> > @relu It looks like this change is causing the error I am seeing.
> > #4
> > What was the issue with not storing failed state?
> > I added some information to this issue in the helm project, as it looks like a change in behavior from v2
> > [helm/helm#7160](https://github.com/helm/helm/issues/7160)
> 
> The issue was that helm would create the "failed" release while it not being stored in the terraform state would cause re-executing apply to fail due to name collision since the provider would not recognize the old failed release and would require manual intervention via helm cli.

okay, that makes sense.

aaronmell:
> I see the tests are failing now, will see if I can fix them. I assume the cause is the updated helm package.

Not failing locally for me, possibly intermittent issue?

relu:
Looks like it, I was able to reproduce it once (exactly when I ran the tests the first time), but never afterwards.

aaronmell:
Can we get someone to rerun CI?
Also need a code review

ncmans:
@aaronmell would be great to update the helm pkg to https://github.com/helm/helm/releases/tag/v3.0.2

aaronmell:
Yeah, ill take a look

shaunc (approved):

Jyoti492:
Hi guys,
any update on the merge requests please ?

aaronmell:
> Hi guys,
> any update on the merge requests please ?

As I said in the Issue, the PR needs reviewed and merged by a maintainer, and a new release cut
@alexsomesan @meyskens @rporres

alex-roman:
Tested the PR today against EKS 1.13 and 1.14 with helm 3.0.1 and 3.0.2 respectively - all good ✅ 
@aaronmell good job man! Thank you!

yuankunzhang:
Folks, can we get this PR merged please?

AndrewDryga:
@meyskens can you please help us with getting this merged?

Jyoti492:
Any news for this to get merged ?

kevinkjt2000:
Is my understanding of helm 3.x correct? Will this PR remove the need for installing tiller onto the cluster?

aaronmell:
> Is my understanding of helm 3.x correct? Will this PR remove the need for installing tiller onto the cluster?

That is correct, tiller no longer exists, all requests use the k8 api now.

lawliet89:
Might be worth posting on the [Google group](https://groups.google.com/forum/#!forum/terraform-tool) to get some eyes from Hashicorp on this. This is what Hashimoto said was the best way to get their attention when I had a chance to ask him. :man_shrugging:

AndrewDryga:
I've dropped a message there https://groups.google.com/forum/#!topic/terraform-tool/URZRQNpoVbM. I was surprised that so many people just leave 👍and did not post it yet 🤷‍♂.

Jyoti492:
I am using the plugin which @aaronmell created for helm 3 
# Cluster RBAC helm Chart repository
data "helm_repository" "cluster_rbac_helm_chart_repo" {
  name = "mcpshareddcr"
  url  = "https://mcpshareddcr.azurecr.io/helm/v1/repo"
  username = var.ARM_CLIENT_ID
  password = var.ARM_CLIENT_SECRET
}

# Deploy Cluster RBAC helm chart onto the cluster
resource "helm_release" "cluster_rbac_helm_chart_release" {
  name  = "mcp-rbac-cluster"
  repository = data.helm_repository.cluster_rbac_helm_chart_repo.metadata[0].name
  chart = "mcp-rbac-cluster"
  version = "0.1.0"
}

# Deploy Namespace RBAC helm chart onto the cluster
resource "helm_release" "namespace_rbac_helm_chart_release" {
  name  = "mcp-rbac-namespace"
  repository = data.helm_repository.cluster_rbac_helm_chart_repo.metadata[0].name
  chart = "mcp-rbac-namespace"
  version = "0.1.0"
}

# Deploy AGIC
resource "helm_release" "agic_helm_chart" {
  name  = "mcp-agic"
  repository = data.helm_repository.cluster_rbac_helm_chart_repo.metadata[0].name
  chart = "mcp-agic"
  version = "1.0.0"
}

but I am facing an issue :
Error: rpc error: code = Unknown desc = configmaps is forbidden: User "system:serviceaccount:kube-system:default" cannot list resource "configmaps" in API group "" in the namespace "kube-system"


Can you please help me what I am doing wrong.

aaronmell:
@Jyoti492 Looks like you don't have permission to deploy to the kube-system namespace. How is your provider configured? Are you trying to deploy a configmap into the kube-system namespace?

Jyoti492:
@aaronmell : my provider is configured as follows
data "azurerm_kubernetes_cluster" "aks_cluster" {
  name                = var.aks_cluster
  resource_group_name = var.resource_group_aks
}

locals {
  kubeconfig_path = "/tmp/kubeconfig"
}

resource "local_file" "kubeconfig" {
  filename = local.kubeconfig_path
  content  = data.azurerm_kubernetes_cluster.aks_cluster.kube_admin_config_raw
}


provider "helm" {
  home = "/home/vsts_azpcontainer/.config/helm/repository"
  kubernetes {
    load_config_file = true
    config_path = local.kubeconfig_path
    config_context = data.azurerm_kubernetes_cluster.aks_cluster.name
  }
}

// Module to deploy Stratus offered helmcharts in AKS cluster
module "mcp_resources" {
  source = "./modules/helm/mcp-resources"
  ARM_CLIENT_ID = var.ARM_CLIENT_ID
  ARM_CLIENT_SECRET = var.ARM_CLIENT_SECRET
  ARM_SUBSCRIPTION_ID = var.ARM_SUBSCRIPTION_ID
  ARM_TENANT_ID = var.ARM_TENANT_ID
}

Jyoti492:
but the service principal which I am using to deploy has cluster admin permissions

aaronmell:
@Jyoti492

Does the context you specify
config_context = data.azurerm_kubernetes_cluster.aks_cluster.name
match the context in kubeconfig_path = "/tmp/kubeconfig"

Does this work?
kubectl --kubeconfig /tmp/kubeconfig -n kube-system --context="Value from data.azurerm_kubernetes_cluster.aks_cluster.name" get cm

Jyoti492:
@aaronmell : what is the best way to provide the cluster config in helm provider?

Jyoti492:
I am trying to output the value of above but the apply doesn't outputs the value.
data "azurerm_kubernetes_cluster" "aks_cluster" {
  name                = var.aks_cluster
  resource_group_name = var.resource_group_aks
}

locals {
  kubeconfig_path = "/tmp/kubeconfig"
}

resource "local_file" "kubeconfig" {
  filename = local.kubeconfig_path
  content  = data.azurerm_kubernetes_cluster.aks_cluster.kube_admin_config_raw
}

output "config_context" {
  value = data.azurerm_kubernetes_cluster.aks_cluster.name
}

output "kubeconfig" {
  value = local.kubeconfig_path
}

Jyoti492:
@aaronmell  : the context is same

aaronmell:
@Jyoti492 I'd check the cluster context in your kubeconfig file. If you only have a single context in the file, you probably don't need it. I don't specify the context in my provider (not using AKS though)

The only thing I specify in the kubernetes section of the helm provider is my config_path

DevOpsFu:
I have an issue with this version of the provider whereby it seems to ignore the `load_config_file` flag being set to `false`. I initialise the provider as follows:

```
provider "helm" {
  kubernetes {
    host                   = module.kubernetesCluster.host
    client_certificate     = base64decode(module.kubernetesCluster.client_certificate)
    client_key             = base64decode(module.kubernetesCluster.client_key)
    cluster_ca_certificate = base64decode(module.kubernetesCluster.cluster_ca_certificate)
    load_config_file       = false
  }
}
```

When applying a chart using this configuration I get the following error:

`Error: invalid configuration: [client-cert-data and client-cert are both specified for clusterUser_containers_devopsfu. client-cert-data will override., client-key-data and client-key are both specified for clusterUser_containers_devopsfu; client-key-data will override, certificate-authority-data and certificate-authority are both specified for devopsfu. certificate-authority-data will override.]`

The cluster mentioned by the error message is in my local kube config file, and unrelated to the cluster I'm targetting when initialising the provider. If I move my `~/.kube/config` file and run it again, it works, so it appears that the provider is always loading the config file regardless of the value of `load_config_file`.

gordonbondon:
`clientcmd.NewNonInteractiveDeferredLoadingClientConfig` was moved outside of `load_config_file` condition https://github.com/terraform-providers/terraform-provider-helm/pull/378/files#diff-cc58f796b951f089e29b5a345f97e5c4R101

This will cause an issue with loading kubeconfig every time. A similar issue was fixed in this provider previously #307

DevOpsFu:
> `clientcmd.NewNonInteractiveDeferredLoadingClientConfig` was moved outside of `load_config_file` condition https://github.com/terraform-providers/terraform-provider-helm/pull/378/files#diff-cc58f796b951f089e29b5a345f97e5c4R101
> 
> This will cause an issue with loading kubeconfig every time. A similar issue was fixed in this provider previously #307

Actually I don't see a call to `NewNonInteractiveDeferredLoadingClientConfig` at all in this branch, other than in the tests. Looks like all of the code that handles the explicit loading of the client config along with any overrides was removed?

helm/provider_test.go
smyrman: Look at https://golang.org/pkg/os/#UserConfigDir

relu:
@DevOpsFu please see my comment [here](https://github.com/terraform-providers/terraform-provider-helm/issues/299#issuecomment-560942206). The issue you described is known.

> > `clientcmd.NewNonInteractiveDeferredLoadingClientConfig` was moved outside of `load_config_file` condition https://github.com/terraform-providers/terraform-provider-helm/pull/378/files#diff-cc58f796b951f089e29b5a345f97e5c4R101
> > This will cause an issue with loading kubeconfig every time. A similar issue was fixed in this provider previously #307
> 
> Actually I don't see a call to `NewNonInteractiveDeferredLoadingClientConfig` at all in this branch, other than in the tests. Looks like all of the code that handles the explicit loading of the client config along with any overrides was removed?

That's because helm v3 uses a different initialization method for the kubernetes client. See [here](https://github.com/terraform-providers/terraform-provider-helm/issues/299#issuecomment-571073667).

helm/provider_test.go
smyrman: Also, chek helm docs -- they may use XDG not only on Linux/BSD, but also on Mac/Windows?

gordonbondon:
@relu looks like a proper solution to avoid hacks (writing private keys and certificates to temporary files, creating empty kubeconfigs etc.) **might** be: 
* get `rest.Config` like implemented here https://github.com/terraform-providers/terraform-provider-kubernetes/pull/690/files#diff-da3a5957d1adf1d97d4dec9f43b36ec1R227 (this PR solves multiple problems in kubernetes provider auth flow)
* Use it to implement `RESTClientGetter.ToRESTConfig` interface https://github.com/helm/helm/blob/master/pkg/action/action.go#L89
* Also just copy `RESTClientGetter.ToDiscoveryClient` https://github.com/kubernetes/cli-runtime/blob/0f3b70f6dbf3547195cf80a6980047a367ed8d0d/pkg/genericclioptions/config_flags.go#L216
* pass this `RESTClientGetter` to `action.Init`

What do you think?

afdezl:
@Jyoti492 I believe when following @aaronmell 's branch you followed the README to compile from the official project master branch and you've built the binary for the latest released version. Clone his repo and switch to the correct branch before compiling. That should solve your issue.

relu:
> @relu looks like a proper solution to avoid hacks (writing private keys and certificates to temporary files, creating empty kubeconfigs etc.) **might** be:
> 
> * get `rest.Config` like implemented here https://github.com/terraform-providers/terraform-provider-kubernetes/pull/690/files#diff-da3a5957d1adf1d97d4dec9f43b36ec1R227 (this PR solves multiple problems in kubernetes provider auth flow)
> * Use it to implement `RESTClientGetter.ToRESTConfig` interface https://github.com/helm/helm/blob/master/pkg/action/action.go#L89
> * Also just copy `RESTClientGetter.ToDiscoveryClient` https://github.com/kubernetes/cli-runtime/blob/0f3b70f6dbf3547195cf80a6980047a367ed8d0d/pkg/genericclioptions/config_flags.go#L216
> * pass this `RESTClientGetter` to `action.Init`
> 
> What do you think?

I would even argue about having this done in helm itself, cause the current auth flow implementation comes from there actually and hence all these limitations.
Might be worth trying out this sort of a bypass solution as you described for the provider itself, even though it comes with a greater complexity cost. I'll probably try this out later today or over the weekend.

davidalger:
As a workaround to `load_config_file` no longer functioning I've got this working with the following provider config:

```

data "aws_eks_cluster" "cluster" {
  name = module.eks.cluster_id
}

data "aws_eks_cluster_auth" "cluster" {
  name = module.eks.cluster_id
}

provider "kubernetes" {
  host                   = data.aws_eks_cluster.cluster.endpoint
  cluster_ca_certificate = base64decode(data.aws_eks_cluster.cluster.certificate_authority.0.data)
  token                  = data.aws_eks_cluster_auth.cluster.token
  load_config_file       = false
  version                = "~> 1.10"
}

provider "helm" {
  kubernetes {
    host                   = data.aws_eks_cluster.cluster.endpoint
    cluster_ca_certificate = base64decode(data.aws_eks_cluster.cluster.certificate_authority.0.data)
    token                  = data.aws_eks_cluster_auth.cluster.token
    config_path            = "/dev/null"
  }
}
```

By specifying `/dev/null` as the path it will no longer pickup whatever is referenced in an exported `KUBECONFIG` env var nor the default kubectl configuration file, which in my case references my `docker-desktop` context and before I added `config_path = "/dev/null"` to the above was giving me the following error:

```
Error: invalid configuration: certificate-authority-data and certificate-authority are both specified for docker-desktop. certificate-authority-data will override.
```

relu:
@gordonbondon here's a branch where I tested your proposed implementation, the code is kind of messy. Unfortunately, I'm having an issue with the release namespace, it seems the kubernetes client is ignoring the provided namespace attribute and all resources are being created in the `default` namespace (even the helm configmaps, which is strange considering that the namespace is fed into the calls directly during driver initialization), looking into ways to overcome this.

https://github.com/relu/terraform-provider-helm/tree/dev-v3-kube-init-refactor

r3kzi:
any news about this?

I am kind of disappointed by the complete lack of communication by the maintainers.

aaronmell:
Looks like a flaky test. @relu Do you mind double checking your stuff to make sure its not that cause of the failure?

aaronmell:
> any news about this?
> 
> I am kind of disappointed by the complete lack of communication by the maintainers.

The project has always been slow to move. But its definitely been more sluggish recently.

relu:
> Looks like a flaky test. @relu Do you mind double checking your stuff to make sure its not that cause of the failure?

My bad, fixing. Need to re-run `go mod vendor` cause some vendored modules changed.

robinkb:
For what it's worth, I've been using your branch to manage my Helm releases, and it works quite well! You have already helped me completely remove Helm 2 from my environments, so thank you! :smile:

Firaenix:
I'm also giving the /aaronmell/terraform-provider-helm/dev-v3 branch a try on Terraform Cloud - in which I'm constantly getting the following error:

```Failed to load config (~/.kube/config): invalid configuration: no configuration has been provided```

Config looks like:

```provider "kubernetes" {
  host  = data.digitalocean_kubernetes_cluster.example.endpoint
  token = data.digitalocean_kubernetes_cluster.example.kube_config[0].token
  cluster_ca_certificate = base64decode(
    data.digitalocean_kubernetes_cluster.example.kube_config[0].cluster_ca_certificate
  )
  load_config_file = false
}

provider "helm" {
  kubernetes {
    host  = data.digitalocean_kubernetes_cluster.example.endpoint
    token = data.digitalocean_kubernetes_cluster.example.kube_config[0].token
    cluster_ca_certificate = base64decode(
      data.digitalocean_kubernetes_cluster.example.kube_config[0].cluster_ca_certificate
    )
    load_config_file = false
    config_path      = "/dev/null"
  }
}
```

Not sure if anyone is aware of this. 
I figured this would be a good place to raise this.

relu:
@Firaenix are you sure you are using the most recent changes? The kubernetes config initialization logic was substantially refactored. I cannot reproduce this currently.

FYI `config_path = "/dev/null"` should no longer be necessary.

Firaenix:
if it's the aaronmell/terraform-provider-helm/dev-v3 branch that is the most up to date, yeah I guess I am.

I also have tried with the config path removed but I get the same issue.

Have you tested terraform cloud?

sagikazarmark:
TBH, I'll be surprised if the maintainers will ever (mean: anytime soon) accept this PR.

Not supporting Helm v2 will not be viable for at least another year.

AFAIR accepting this provider into the official list of providers was a painfully slow process and it looks like development hasn't changed either.

All in all: I think a maintaining a community fork at this point is justified. Thanks to the new [terraform registry](https://www.hashicorp.com/blog/announcing-providers-in-the-new-terraform-registry) we should soon be able to publish providers (eg. `helm3` ?) ourselves to the registry, making adoption easier.

jrhouston:
👋 

Thanks so much for taking on this feature @aaronmell. 

I've just joined the providers team at HashiCorp and I'm going to be one of the maintainers for the Helm provider going forward. I wanted to drop a reply here to say I'm very keen to get v3 in, and that we should be seeing a bit more traction on this repository. We are currently a very small team so we're super appreciative of your contributions, and your patience in waiting for this feature to land. 

@aaronmell Are you planning on adding anything more to this PR, or is it ready for review?

aaronmell:
@jrhouston I have nothing else to contribute. I know that @relu was working on a few things, but I believe they are optional

relu:
Great hearing the news, @jrhouston!

@aaronmell that's right, the import feature I've worked on can be introduced in a future release.

I was not able to @Firaenix in a "sterile" environment using the same auth method, though I admit I was not yet able to test it on TF cloud unfortunately. I think some more testing might be necessary and maybe some tidying up here and there, but aside from that it should be good to go.

wstewartii:
Is this under review? I've had success using it and would like to see it merged into the official helm provider.

SubatomicHero:
Cant wait for this to be merged in. Managing v2 with TLS via Terraform is a pain! Also, can we please have good clear examples of how to configure the provider with various configurations e.g. GKE, EKS. Thank you!

jakesylvestre:
By the way, I think #359 might be a prerequisite here

Also @sagikazarmark sounds great!

ghost:
Heya @jakesyl isn't helm 3 "tiller-less"? Why should one need to issue a helm install in the first place? Or am I missing something?

jakesylvestre:
Yeah- the absence of tiller is why it’s a prerequisite. Tiller is no longer in the cluster which means every operation required a port forward since they’re running outside of the cluster

relu:
@jakesyl tiller is no longer in the cluster because it does not exist anymore, hence there's nothing to port-forward to. The current implementation of Helm v3 uses configmaps or secrets to store release metadata and Helm interacts with Kubernetes directly through the api client. That being said, I believe the issue you mentioned does not apply in our case here.

wstewartii:
> Cant wait for this to be merged in. Managing v2 with TLS via Terraform is a pain! Also, can we please have good clear examples of how to configure the provider with various configurations e.g. GKE, EKS. Thank you!

https://github.com/terraform-providers/terraform-provider-helm/issues/299#issuecomment-576830992


```
git clone https://github.com/terraform-providers/terraform-provider-helm.git
cd terraform-provider-helm
git fetch origin +refs/pull/378/merge
git checkout FETCH_HEAD
make build
mkdir -p $HOME/terraform.d/plugins
cp terraform-provider-helm $HOME/.terraform.d/plugins/terraform-provider-helm
```

You'll need  make, git and golang version 1.12 or greater to build the provider

rigratz:
@aaronmell This is awesome, can't wait for this to get merged in. Do you have any recommendation for cleanly migrating an existing helm2 release within Terraform? I haven't been successful yet doing anything other than deploying with this from a clean state. Which isn't the end of the world, but I always prefer to minimize downtime whenever possible.

aaronmell:
@rigratz I haven't tried upgrading. I believe there is a helm plugin for that, but I haven't looked into it.

rigratz:
Oh yeah, the plugin works great for a manual migration of a release. But the next time I run my TF pipeline after the migration it tries to 'create' again and hits a conflict because a release with the same name exists but the state doesn't get updated accordingly (since Terraform didn't handle the migration). Probably just a TF state quirk so was hoping maybe you'd run into it before. No worries, thanks for the response!

relu:
@rigratz you can try out this not very well tested import feature support implementation, it worked fine for my use cases: https://github.com/aaronmell/terraform-provider-helm/pull/7

jclynny:
Excited to see progress on this, I hope this becomes available soon!

jrhouston (approved):

bartlett-ops:
:tada:

jrhouston:
Huge thank you to @aaronmell and @relu for this contribution and to all who reviewed and commented. ❤️

jonathan-mothership:
Best merge I've seen all year! Thanks to all the contributors!

donaldguy:
Now the operative question is what the expectation is on a release timeline ? (“at least” to registry) I would love to get the loading of these binaries from a git submodule out of my TF Cloud runs (would make speculative plans way more usable again)

jrhouston:
@donaldguy The release is running now. 😄

inyono:
Thanks for the awesome work! Upgrading was rather easy. To migrate, I did:
- Update the provider version
- [Install the Helm-2to3-Plugin](https://helm.sh/blog/migrate-from-helm-v2-to-helm-v3/#helm-2to3-plugin)
- [Migrate Helm v2 releases](https://helm.sh/blog/migrate-from-helm-v2-to-helm-v3/#migrate-helm-v2-releases)
- Do a `terraform apply` (note that this should only contain `changes`, no more `additions`)
- [Clean up of Helm v2 data](https://helm.sh/blog/migrate-from-helm-v2-to-helm-v3/#clean-up-of-helm-v2-data)

PR #383, opened 2019-12-16
Update Github issue template

paaloeye:
I found `helm_resource` is quite confusing, so I think it's a good idea to actualise it.


meyskens (approved):
Good spot! Thanks

PR #391, opened 2020-02-04
Fix test infra on GKE

jrhouston:
This fixes the following error when trying to spin up the test infrastructure:

```
Error: node_version and min_master_version must be set to equivalent values on create
```


alexsomesan (approved):
LGTM!
Thanks, @jrhouston’

PR #394, opened 2020-02-05
Helm v3 import release feature

relu:
This adds support for importing helm releases into the terraform state.

Example use case:
```terraform
data "helm_repository" "stable" {
  name = "stable"
}

  name        = "mariadb-example"
  repository  = data.helm_repository.stable.metadata.0.name
  chart       = "mariadb"
  namespace   = "example"
  max_history = 5
  timeout     = 60
}
```
```
$ terraform import helm_release.example example/mariadb-example
```

The implementation logic is definitely open for debate.

TODO:
- [x] Test coverage
- [x] Documentation
- [ ] Bump version


helm/resource_release.go
```diff
@@ -776,3 +783,40 @@ func chartPathOptions(d resourceGetter, m *Meta) (*action.ChartPathOptions, stri
...
+	d.Set("name", r.Name)
+
+	if err := setIDAndMetadataFromRelease(d, r); err != nil {
+		return nil, err
+	}
+
+	return schema.ImportStatePassthrough(d, meta)
+}
+
+func idParts(id string) (string, string, error) {
```
jrhouston: I'd like to change this to a more explicit name like `parseImportIdentifier` since it's only used when importing, and does the pretty explicit job of parsing the Id string into a namespace and name. I think "Parts" is a detail of the function's implementation.

helm/resource_release.go
```diff
@@ -776,3 +783,40 @@ func chartPathOptions(d resourceGetter, m *Meta) (*action.ChartPathOptions, stri
 		Password: d.Get("repository_password").(string),
 	}, chartName, nil
 }
+
+func resourceHelmReleaseImportState(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
```
jrhouston, line 870: Something that we're not doing here is accounting for the default values that can't be imported from the helm data. We'll need a way to set the defaults, if they're not set in the HCL, when the resource is being imported.

For example, using this config:
```hcl
provider "helm" {}

data "helm_repository" "stable" {
  name = "stable"
  url = "https://kubernetes-charts.storage.googleapis.com/"
}

resource "helm_release" "mysql" {
  name = "mysql-test-db"
  repository = data.helm_repository.stable.metadata[0].name
  chart = "mysql"
}
```

When I run `terraform import default/mysql-test-db` it successfully imports. However when I run plan again, this happens:

```
  # helm_release.mysql will be updated in-place
  ~ resource "helm_release" "mysql" {
      + atomic                = false
        chart                 = "mysql"
      + cleanup_on_fail       = false
      + dependency_update     = false
      + disable_crd_hooks     = false
      + disable_webhooks      = false
      + force_update          = false
        id                    = "mysql-test-db"
      + max_history           = 0
        metadata              = [
            {
                chart     = "mysql"
                name      = "mysql-test-db"
                namespace = "default"
                revision  = 1
                values    = "null"
                version   = "1.6.2"
            },
        ]
        name                  = "mysql-test-db"
        namespace             = "default"
      + recreate_pods         = false
      + render_subchart_notes = true
      + replace               = false
      + repository            = "stable"
      + reset_values          = false
      + reuse_values          = false
      + skip_crds             = false
        status                = "deployed"
      + timeout               = 300
      + verify                = false
        version               = "1.6.2"
      + wait                  = true
    }

Plan: 0 to add, 1 to change, 0 to destroy.

------------------------------------------------------------------------
```

helm/resource_release_test.go
```diff
@@ -57,6 +57,33 @@ func TestAccResourceRelease_basic(t *testing.T) {
...
+	name := fmt.Sprintf("test-import-%s", acctest.RandString(10))
+	namespace := fmt.Sprintf("%s-%s", testNamespace, acctest.RandString(10))
+	// Delete namespace automatically created by helm after checks
+	defer deleteNamespace(t, namespace)
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:     func() { testAccPreCheck(t, namespace) },
+		Providers:    testAccProviders,
+		CheckDestroy: testAccCheckHelmReleaseDestroy(namespace),
+		Steps: []resource.TestStep{{
```
jrhouston, line 70: Following on from my comment [here](https://github.com/terraform-providers/terraform-provider-helm/pull/394/files#r375605806). I would propose adding some additional checks that cover the case where we check for default values that haven't been set in the HCL.

jrhouston (changes requested):
This looks great so far, thank you for contributing @relu! If you need any help with the tests or updating the documentation let me know. 😄

helm/resource_release.go
relu: Yes, I'm aware of this issue and was planning to figure it out soon enough. I was thinking the only reasonable way to do this was to create a "default values" map that would be referenced both from the resource schema definition and the import function.

helm/resource_release_test.go
relu: This test is currently failing and I'm in a bit of a deadlock right now cause I can't really figure out how to make it work.

Let me know if you have any ideas

relu:
I noticed the acceptance tests job no longer ran in Travis, was it disabled?

helm/resource_release_test.go
relu: Think I've figured it out :)

helm/resource_release.go
relu: Please check my solution and let me know what you think. I've managed to get all the defaults in except for the `repository` attribute which we have absolutely no information about from the helm release itself (looks like it never gets stored as part of the release metadata, might make sense to raise this to the helm maintainers).

jrhouston:
> I noticed the acceptance tests job no longer ran in Travis, was it disabled?

Looks like there was some issues with `.travis.yaml`. I've opened #407 to address.

helm/resource_release_test.go
```diff
@@ -56,6 +55,51 @@ func TestAccResourceRelease_basic(t *testing.T) {
...
+				func() []resource.TestCheckFunc {
+					importTestCheckFuncs := []resource.TestCheckFunc{
+						resource.TestCheckResourceAttr("helm_release.imported", "metadata.0.revision", "1"),
+						resource.TestCheckResourceAttr("helm_release.imported", "metadata.0.version", "7.1.0"),
+						resource.TestCheckResourceAttr("helm_release.imported", "status", release.StatusDeployed.String()),
+						resource.TestCheckNoResourceAttr("helm_release.imported", "repository"),
+					}
+
+					// Test for default values being set
+					for key, value := range DefaultAttributes {
```
jrhouston: Further to my comment above, I think using this variable means this acceptance test is relying upon an implementation detail of the provider. It would be better if we wrote out the checks explicitly. 

It's tedious, but it means at a glance you can look at the test and know exactly what is being tested.

helm/resource_release.go
```diff
@@ -24,13 +24,36 @@ import (
 // errReleaseNotFound is the error when a Helm release is not found
 var errReleaseNotFound = errors.New("release not found")
 
+// DefaultAttributes release attribute values
+var DefaultAttributes = map[string]interface{}{
```
jrhouston: This should probably be private like the variable above.

jrhouston (changes requested):

jrhouston:
Requested a couple of changes but otherwise this is great work thank you @relu.

helm/resource_release_test.go
relu: I agree with your assessment.

jrhouston:
I'm seeing errors when running the acceptance tests in this branch. The issue looks to be with the `description` field but I haven't investigated further yet.

```
--- FAIL: TestAccResourceRelease_postrender (60.12s)
    provider_test.go:148: [DEBUG] Creating namespace terraform-acc-test-40t6ntjnfg
    testing.go:654: Step 0 error: After applying this step, the plan was not empty:

        DIFF:

        UPDATE: helm_release.test
          atomic:                     "false" => "false"
          chart:                      "mariadb" => "mariadb"
          cleanup_on_fail:            "false" => "false"
          dependency_update:          "false" => "false"
          description:                "Install complete" => ""
          disable_crd_hooks:          "false" => "false"
          disable_openapi_validation: "false" => "false"
          disable_webhooks:           "false" => "false"
          force_update:               "false" => "false"
          id:                         "test" => "test"
          max_history:                "0" => "0"
          metadata.#:                 "1" => "1"
          metadata.0.chart:           "mariadb" => "mariadb"
          metadata.0.name:            "test" => "test"
          metadata.0.namespace:       "terraform-acc-test-40t6ntjnfg" => "terraform-acc-test-40t6ntjnfg"
          metadata.0.revision:        "1" => "1"
          metadata.0.values:          "{\"master\":{\"persistence\":{\"enabled\":false}},\"replication\":{\"enabled\":false}}" => "{\"master\":{\"persistence\":{\"enabled\":false}},\"replication\":{\"enabled\":false}}"
          metadata.0.version:         "7.1.0" => "7.1.0"
          name:                       "test" => "test"
          namespace:                  "terraform-acc-test-40t6ntjnfg" => "terraform-acc-test-40t6ntjnfg"
          postrender.#:               "1" => "1"
          postrender.0.binary_path:   "cat" => "cat"
          recreate_pods:              "false" => "false"
          render_subchart_notes:      "true" => "true"
          replace:                    "false" => "false"
          repository:                 "https://kubernetes-charts.storage.googleapis.com" => "https://kubernetes-charts.storage.googleapis.com"
          reset_values:               "false" => "false"
          reuse_values:               "false" => "false"
          set.#:                      "2" => "2"
          set.0.name:                 "master.persistence.enabled" => "master.persistence.enabled"
          set.0.value:                "false" => "false"
          set.1.name:                 "replication.enabled" => "replication.enabled"
          set.1.value:                "false" => "false"
          set_sensitive.#:            "0" => "0"
          set_string.#:               "0" => "0"
          skip_crds:                  "false" => "false"
          status:                     "deployed" => "deployed"
          timeout:                    "300" => "300"
          verify:                     "false" => "false"
          version:                    "7.1.0" => "7.1.0"
          wait:                       "true" => "true"
```

jrhouston (approved):

jrhouston:
Thank you @relu!

aristosvo:
@jrhouston Any idea when we can expect the v1.0.1 release with this feature?

bigwheel:
It seems like 1.1.0 contains this feature.
https://github.com/terraform-providers/terraform-provider-helm/blob/080e3d99/CHANGELOG.md#110-unreleased
I'm so haapy because I'm migrating helm install script to terraform tf files just now.
Many thanks.

PR #407, opened 2020-02-12
Use kind to run acceptance tests in travis-ci

jrhouston:


relu:
I'll send you a fix for the failing test, there's a race condition caused by the repo resource name collision.

jrhouston:
Awesome, thanks @relu.

jrhouston:
Hmm, looks like there's still a flakey test.

relu:
Indeed, and this one doesn't seem to be caused by parallelization; I wasn't able to reproduce it locally. I suspect it has something to do with the release resource being initialized before the data repo is updated, but have no idea right now how to investigate that further.

relu:
Would be great to get this into `master` so other PRs can benefit from functioning tests.

jrhouston:
> Would be great to get this into `master` so other PRs can benefit from functioning tests.

Just want to see if I can fix this without using `travis_retry` as that's going to add a fair bit of time to the build in the worst case.

relu:
In an ideal scenario, these tests should rely on fixtures, mocking actual requests to chart repositories so that any potential network blips would not affect the outcome.

Would be great if something like [this](https://github.com/seborama/govcr) could be used, but I don't see how that would be possible considering that helm doesn't allow overriding the [ChartDownloader](https://pkg.go.dev/helm.sh/helm/v3@v3.1.1/pkg/downloader?tab=doc#ChartDownloader) `Getters` ([details](https://github.com/helm/helm/blob/v3.1.1/pkg/action/install.go#L711-L720)).

jrhouston:
I think I finally figured this out. A temporary directory for the helm config files was being created for every individual test run, and then the paths were supplied to helm using `os.Setenv()`. This is problematic because the tests are run in parallel but share the same env.

For the moment I have made a change to create just one temp dir every time the whole test suite runs.   This works because the repository data source creates a lock before it does anything. I've run the test suite a bunch of times and this seems to have abated the flakiness for now.

I didn't see an easy way to supply the paths directly to helm on each individual test run. However, I'm anticipating changes to the terraform-plugin-sdk's testing framework that will make testing with this level of isolation much easier so I don't want to spend too much time investigating this now.

relu:
Now it makes a lot of sense, great finding! I think it should be fine as long as each test case uses its own unique name for repositories which should be the case now after previous changes.

relu (approved):

jrhouston:
Thanks for your help on this @relu :bowing_man:

PR #439, opened 2020-03-12
Make helm_resource.description a computed field

jrhouston:
After looking into this field some more, it seems that it is set by the helm client itself. There doesn't seem to be a way for an operator to set this field using the `helm` command. e.g [here](https://github.com/helm/helm/blob/357d265de0e72f02c493701cdbee9ecc6eb0a016/pkg/action/uninstall.go#L95)

@relu FYI.

### DIFF
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -300,8 +300,8 @@ func resourceRelease() *schema.Resource {
 			},
 			"description": {
 				Type:        schema.TypeString,
-				Optional:    true,
-				Description: "Add a custom description",
+				Computed:    true,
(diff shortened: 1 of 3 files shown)


relu:
The Helm [changelog](https://github.com/helm/helm/releases/tag/v3.1.0) contradicts this.

> The --description flag, present in Helm v2, was restored to v3. This flad enables custom release descriptions

Feature PR: https://github.com/helm/helm/pull/7074

jrhouston:
🤦‍♂ Totally missed that, thanks.

PR #442, opened 2020-03-15
Example with chart not in repository

borgoat:
Hey there!

It took me longer than I'd like to admit to figure out that the Helm provider allows to use a local chart and not just those published in a repo...
So I figured it'd be good to have this documented!

It all started when I wanted to install [Vault via Helm](https://github.com/hashicorp/vault-helm) where there has been an issue since a while asking to publish the chart in a public repo... hashicorp/vault-helm#14 
Of course it'd still be better if that was available in a repository, but it's still good to be able to point to local charts, especially while developing them

### DIFF
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -99,7 +99,7 @@ func resourceRelease() *schema.Resource {
 			"chart": {
 				Type:        schema.TypeString,
 				Required:    true,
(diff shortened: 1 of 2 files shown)


jrhouston:
Nice! Thank you @giorgioazzinnaro

jrhouston (approved):

PR #447, opened 2020-03-18
Release v1.1.0

jrhouston:
* Update CHANGELOG
* Fix typo in skip_crds field description
* Suppress diff when description attribute is not set

### DIFF
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -261,7 +261,7 @@ func resourceRelease() *schema.Resource {
 				Type:        schema.TypeBool,
 				Optional:    true,
 				Default:     defaultAttributes["skip_crds"],
-				Description: "If set, no CRDs will be installed. By DefaultAttributes, CRDs are installed if not already present",
+				Description: "If set, no CRDs will be installed. By default, CRDs are installed if not already present",
 			},
 			"render_subchart_notes": {
 				Type:        schema.TypeBool,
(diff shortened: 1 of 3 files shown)


helm/resource_release.go
```diff
@@ -302,6 +302,9 @@ func resourceRelease() *schema.Resource {
 				Type:        schema.TypeString,
 				Optional:    true,
 				Description: "Add a custom description",
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
```
jrhouston, line 305: I have suppressed this because when we import a release and don't set a description terraform will render the diff changing `Install complete` to `null`. I have made the assumption here that if the user hasn't set a custom description field then they probably don't care to capture the default one that gets set by helm.

jrhouston:
@dak1n1 I have added you as a reviewer just for your information.

helm/resource_release.go
alexsomesan: Would setting this to `Computed` also have worked? That's sort of the canonical way to treat such cases.
Regardless, I'm with you that Description is likely not useful in any non-human context so not capturing in state should not make much of a difference.

alexsomesan (approved):
LGTM

helm/resource_release.go
jrhouston: That's what I thought initially, but they added the `--description` feature to the helm CLI so there are potentially situations where the operator will want to explicitly set this field. See #439

PR #449, opened 2020-03-20
Fix repo/chart path bug causing unexpected diff

### LINKED ISSUE #448: Helm provider 1.1 constantly updating helm_release
# Terraform Version
0.12.24

### Affected Resource(s)
- helm_release

```hcl
resource "helm_release" "cert_manager" {
  name      = "cert-manager"
  chart     = "jetstack/cert-manager"
  namespace = kubernetes_namespace.cert_manager.id
  version   = "v0.13.0"
```

### Expected Behavior
No changes after chart has been applied.

### Actual Behavior
Every time chart is read as `cert-manager`. Same for `stable/*` charts.
This only happens in newly released 1.1 version.

```
  # helm_release.cert_manager will be updated in-place
  ~ resource "helm_release" "cert_manager" {
        atomic                     = false
      ~ chart                      = "cert-manager" -> "jetstack/cert-manager"
```

jrhouston:
Resolves #448

### DIFF
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -628,9 +628,6 @@ func resourceDiff(d *schema.ResourceDiff, meta interface{}) error {
 
 func setIDAndMetadataFromRelease(d *schema.ResourceData, r *release.Release) error {
 	d.SetId(r.Name)
-	if err := d.Set("chart", r.Chart.Metadata.Name); err != nil {
-		return err
-	}
 
 	if err := d.Set("version", r.Chart.Metadata.Version); err != nil {
 		return err
@@ -890,6 +887,7 @@ func resourceHelmReleaseImportState(d *schema.ResourceData, meta interface{}) ([
 
 	d.Set("name", r.Name)
 	d.Set("description", r.Info.Description)
+	d.Set("chart", r.Chart.Metadata.Name)
 
 	for key, value := range defaultAttributes {
 		d.Set(key, value)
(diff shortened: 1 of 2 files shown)


helm/resource_release.go
```diff
@@ -890,6 +887,7 @@ func resourceHelmReleaseImportState(d *schema.ResourceData, meta interface{}) ([
 
 	d.Set("name", r.Name)
 	d.Set("description", r.Info.Description)
+	d.Set("chart", r.Chart.Metadata.Name)
```
dak1n1, line 890: I ran a lint tool on this and it mentioned the error return value isn't being checked for this one. It could be fine. But I'll post the output here just in case:

```
[dakini@dax terraform-provider-helm ((cbb5efb...))]$ golangci-lint run
helm/resource_release.go:888:7: Error return value of `d.Set` is not checked (errcheck)
        d.Set("name", r.Name)
             ^
helm/resource_release.go:889:7: Error return value of `d.Set` is not checked (errcheck)
        d.Set("description", r.Info.Description)
             ^
helm/resource_release.go:890:7: Error return value of `d.Set` is not checked (errcheck)
        d.Set("chart", r.Chart.Metadata.Name)
             ^
helm/resource_release.go:831:5: S1003: should use !strings.Contains(name, "/") instead (gosimple)
        if strings.Index(name, "/") == -1 && repository != "" {
           ^
helm/provider.go:462:27: S1019: should use make([]string, len(s)) instead (gosimple)
        result := make([]string, len(s), len(s))
```

jrhouston: Yeah - there is a bunch of linter problems with this repo but it shouldn't be a problem for this particular bug. We should totally clean these up though.

dak1n1 (approved):
LGTM

PR #464, opened 2020-04-10
helm_release: chart lint when `terraform plan`

mcuadros:
### Description

This introduces the equivalent to the execution of `helm lint` in the `terraform plan` since currently errors with malformed values or malformed remote/local charts are only found in the `terraform apply` 

### Acceptance tests

### DIFF
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -617,13 +617,21 @@ func resourceDiff(d *schema.ResourceDiff, meta interface{}) error {
 		return nil
 	}
 
+	// Validates the resource configuration, the values, the chart itself, and
+	// the combination of both.
+	//
+	// Maybe here is not the most canonical place to include a validation
+	// but is the only place to fail in `terraform plan`.
+	if err := resourceReleaseValidate(d, meta.(*Meta), cpo); err != nil {
+		return err
+	}
+
 	// Set desired version from the Chart metadata if available
 	if len(c.Metadata.Version) > 0 {
 		return d.SetNew("version", c.Metadata.Version)
-	} else {
-		return d.SetNewComputed("version")
 	}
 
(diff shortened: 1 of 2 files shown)


mcuadros:
Supersets https://github.com/terraform-providers/terraform-provider-helm/pull/283

helm/resource_release.go
```diff
@@ -617,13 +617,21 @@ func resourceDiff(d *schema.ResourceDiff, meta interface{}) error {
 		return nil
 	}
 
+	// Validates the resource configuration, the values, the chart itself, and
+	// the combination of both.
+	//
+	// Maybe here is not the most canonical place to include a validation
```
jrhouston, line 623: I agree this is not the canonical place to put this. It would be better to put this in a `ValidateFunc` on the `chart` attribute, but there we don't have access to the provider meta struct so we wouldn't be able to validate with the supplied values.

mcuadros: Yep, this was the only method I was able to make it work in the plan stage.

jrhouston:
I think if we're going to add this functionality we should expose an attribute that allows the user to:
- Enable/disable linting
- Toggle linting strictness (i.e abort the plan on warnings)

I would propose that by default lint is enabled and strict mode is disabled. What do you think @mcuadros?

mcuadros:
Humm, why you want to apply something that is going to fail? The linting errors aborting a plan, are always failed applies. 

In other words, any failed plan will fail always to apply. This is not style linting or something like this, is about template compilations errors.

jrhouston:
Ah, misunderstood the intent of this. We're just trying to catch invalid YAML at plan time, not linting per say.

jrhouston (approved):

Issue #251, opened 2019-04-12
Sensitive Values leak when changed/removed


rmb938:
### Terraform Version
```
Terraform v0.11.7
```

### Affected Resource(s)

- helm_release

### Terraform Configuration Files
```hcl
resource "helm_release" "thanos" {
  name       = "thanos"
  repository = "${var.helm_repo}"
  chart      = "thanos"
  version    = "0.2.0"
  namespace  = "${module.prometheus.namespace}"

  set_sensitive {
    name  = "s3.awsAccessKey"
    value = "access key"
  }
  set_sensitive {
    name  = "s3.awsSecretKey"
    value = "secret key"
  }

}
```

### Expected Behavior

Diff not to show sensitive values

### Actual Behavior

Diff showed sensitive values

```
      set_sensitive.#:                "2" => "0"
      set_sensitive.3469752628.name:  "s3.awsSecretKey" => ""
      set_sensitive.3469752628.value: "some_secret_here" => ""
      set_sensitive.4209480927.name:  "s3.awsAccessKey" => ""
set_sensitive.4209480927.value: "other_secret_here" => ""
```

### Steps to Reproduce
1. `terraform apply`
1. Remove `set_sensitive` values
1. `terraform plan`

### Important Factoids

This seems like it would be a terraform bug instead of one in the provider but I figured I would start here first.

dineshba:
As @rmb938 you mentioned it is `terraform` bug [refer](https://github.com/hashicorp/terraform/issues/20051).

gordonbondon:
I don't know if I should create another issue. But with terraform > 0.12 helm provider leaks secrets when anything in release changes because all of the unchanged attributes are also shown. And they include `metadata = [ { values ="..." } ]` which displays all of the values in plaintext during plan/apply.

bjaworski3:
Is this being worked on right now? With terraform > 0.12, we basically can't use the helm provider because any changes in the helm chart will print out all values in plaintext via the `metadata.values` field, including any `set_sensitive` values

aareet:
Closing in favour of https://github.com/terraform-providers/terraform-provider-helm/pull/480

PR #473, opened 2020-04-21
Metadata should be marked sensitive as it can contain secrets

jbehling:
### Description
All values from a helm release are stored in the metadata field. We can mask secrets with `set_sensitive`, but once the values make it into the metadata, they will leak on a diff. Since there is no way to mark an attribute as conditionally sensitive in a terraform schema, we should always assume this field is sensitive.  

Losing the metadata field from diffs, doesn't have any consequence that I can think of. If you are changing the inputs to a release, then those changes are reflected via the  `values`, `set`, and `set_sensitive` blocks. 

### Acceptance tests

### References
https://github.com/terraform-providers/terraform-provider-helm/issues/251

### DIFF
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -325,6 +325,7 @@ func resourceRelease() *schema.Resource {
 				Type:        schema.TypeList,
 				Computed:    true,
 				Description: "Status of the deployed release.",
+				Sensitive:   true,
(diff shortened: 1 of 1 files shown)


aareet:
Closing in favour of https://github.com/terraform-providers/terraform-provider-helm/pull/480

PR #477, opened 2020-04-22
Helm v3.2.0

relu:
### Description

Helm v3.2.0 was released and this PR implements its new added features into the provider.

### Acceptance tests

https://gist.github.com/relu/942bc572169483c6a36ca85ea57e374f#file-testacc-log

### References

Helm Release: https://github.com/helm/helm/releases/tag/v3.2.0
Features:
- create_namespace attribute (release): https://github.com/helm/helm/commit/694cf61aacba6bc4ed59a65d23acd20f6c8b95bd
- skip_crds on upgrade (release): https://github.com/helm/helm/commit/e92a258a9d7cc684589cb22c317eb7ddaeaf753e
- sql helm_driver support (release): https://github.com/helm/helm/commit/21d2aa7f2b28e0ecec0cd17825a422a692c71752


website/docs/index.html.markdown
```diff
@@ -88,7 +88,7 @@ The following arguments are supported:
 * `registry_config_path` - (Optional) The path to the registry config file. Defaults to `HELM_REGISTRY_CONFIG` env if it is set, otherwise uses the default path set by helm.
 * `repository_config_path` - (Optional) The path to the file containing repository names and URLs. Defaults to `HELM_REPOSITORY_CONFIG` env if it is set, otherwise uses the default path set by helm.
 * `repository_cache` - (Optional) The path to the file containing cached repository indexes. Defaults to `HELM_REPOSITORY_CACHE` env if it is set, otherwise uses the default path set by helm.
-* `helm_driver` - (Optional) "The backend storage driver. Valid values are: `configmap`, `secret`, `memory`. Defaults to `secret`
+* `helm_driver` - (Optional) "The backend storage driver. Valid values are: `configmap`, `secret`, `memory`, `sql`. Defaults to `secret`
```
jrhouston: Maybe we add more detail to the docs on how to configure this? 

What about an additional block for configuring the sql dialect & connection string?

helm/structure_kubeconfig.go
```diff
@@ -0,0 +1,177 @@
+package helm
```
jrhouston, line 1: Thanks for breaking this out

jrhouston (changes requested):
This is great, thanks for doing this @relu - just a couple of comments about the sql backend.

website/docs/index.html.markdown
relu: It doesn't seem to support anything other than [postgres for now](https://github.com/helm/helm/blob/master/pkg/storage/driver/sql.go#L184) and the way the connection string can be configured is through an environment variable `HELM_DRIVER_SQL_CONNECTION_STRING`. I'll add a note about that in the docs.
I actually had mixed feelings about even adding support for this feature in the provider at this stage, but in the end, I figured it might come in handy for some folks and it might also help out with testing since it's considered a beta.

jrhouston (approved):

helm/data_repository.go
```diff
@@ -40,6 +40,11 @@ func dataRepository() *schema.Resource {
 				Optional:    true,
 				Description: "Verify certificates of HTTPS-enabled servers using this CA bundle",
 			},
+			"insecure_skip_tls_verify": {
```
mcuadros: Adding, new configuration to a deprecated resource maybe it's not a good idea.

relu: I agree though I have a question. What are the plans for supporting helm repo advanced features in the future if not through a data resource? I'm thinking mostly about authentication which is most likely something folks use.

mcuadros: The `repository` is pojo that contains that used to retrieve the package, all their options were available in the release call. But this one looks like it's [not available](https://github.com/helm/helm/blob/master/pkg/action/install.go#L114), since only affects to the index requests. 

In this case, since we are removing `repository` no one needs to download the index with an invalid certificate. :rofl:

relu: Clear now, missed the fact you can already configure these though the release resource. The only inconvenience I see is when dealing with multiple releases from the same repo, there's going to be a lot of duplication, but that's going to be a separate conversation.

Should I revert the change? I actually didn't realize the repo data resource was marked deprecated before starting to work on my PR :)

mcuadros: Yes, please remove this. We are not doing more updates on this resource. You can use vars to avoid duplication. `repository` has several problems but the main one is that is persisted on the client-side, causing a lot of pain to everyone.

PR #340, opened 2019-09-19
metadata values can sometimes be sensitive due to the usage of set_sensitive

rmb938:
Fixes #251 
Fixes #333

WIP

This doesn't seem to be working on existing resources... weird.

PR #480, opened 2020-04-26
helm_release: cloak metadata.value and deprecate set_string

### LINKED ISSUE #424: Metadata output contains sensitive values
### Terraform Version

0.12.17

### Affected Resource(s)

helm_release

```hcl
resource "helm_release" "my-release" {
  ...  
  values       = [file("values.yaml")]

  set {
    name  = "image.version"
    value = file("app.version")
  }

  set_sensitive {
    name  = "variables.secrets.DB_PASSWORD"
    value = "pa$$word"
  }
}
```
### Expected Behavior

when I do terrafrom apply it shows update plan with changes. In that plan metadata output is exposed and that metadata contains previous values and additional sets. I expect that sensitive keys are somehow obfuscated and don't show in the logs.

### Actual Behavior

The problem is that sensitive values are in plain text.
helm_release.hello-cloud will be updated in-place
```hcl
  ~ resource "helm_release" "my-release" {
        ....
        metadata              = [
            {
                ...
                values    = jsonencode(
                    {
                            ........
                            secrets = {
                                DB_PASSWORD = "pa$$word"
                            }
                        }
                    }
                )
        ]

       set_sensitive {
            name  = "variables.secrets.DB_PASSWORD"
            value = (sensitive value)
        }
....
}
```
### Steps to Reproduce

1. `terraform apply`

mcuadros:
#### Sensitive data leacks

The problems exposed at #424 are fixed cloaking the values the `metadata.0.values` searching the values defined by `set_sensitive` arguments and replacing it with `(sensistive_value)`. 

```      
     metadata = [
            {
                chart     = "podinfo"
                name      = "podinfo"
                namespace = "test-providers"
                revision  = 1
                values    = jsonencode(
                    {
                        replicaCount = 1
                        service      = {
                            password = "(sensitive value)"
                        }
                    }
                )
                version   = "3.2.2"
            },
     ]
```

This is replaced inside of the provider, and not just simply turning the `values` attribute into a sensitive value. I tried to add a `values_raw` with the original data and mark this as sensitive but didn't work, it looks like terraform was ignoring this and just displaying the values. 

####  `set_string` deprecation

`set_string` it's a very ad-hoc argument, implemented in a way that simply doesn't scale. `set_sensitive` was missing the functionality of `string` so following the pattern we should introduce `set_sensitive_string` what it's simply horrible.

Instead of doing this `set_string` has been deprecated in favor of `set` with the option `type` equal to `string`, behaves exactly as `set_string`. Additionally now `set_sensitive` also has the functionally as string. 

```
  set_sensitive {
    name  = "very-secret-number"
    value = 42
    type = "string"
  }
```

### Acceptance tests

### References
Fixes #424 
Closes #473
Closes #340

### DIFF
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
 	"github.com/pkg/errors"
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/chart"
@@ -19,7 +20,6 @@ import (
 	"helm.sh/helm/v3/pkg/postrender"
 	"helm.sh/helm/v3/pkg/release"
 	"helm.sh/helm/v3/pkg/strvals"
-
 	"sigs.k8s.io/yaml"
 )
 
@@ -136,6 +136,13 @@ func resourceRelease() *schema.Resource {
 							Type:     schema.TypeString,
 							Required: true,
 						},
+						"type": {
+							Type:     schema.TypeString,
+							Optional: true,
+							ValidateFunc: validation.StringInSlice([]string{
+								"auto", "string",
+							}, false),
+						},
 					},
 				},
 			},
@@ -154,6 +161,13 @@ func resourceRelease() *schema.Resource {
 							Required:  true,
 							Sensitive: true,
 						},
+						"type": {
+							Type:     schema.TypeString,
+							Optional: true,
+							ValidateFunc: validation.StringInSlice([]string{
+								"auto", "string",
+							}, false),
+						},
 					},
 				},
 			},
@@ -161,6 +175,8 @@ func resourceRelease() *schema.Resource {
 				Type:        schema.TypeSet,
 				Optional:    true,
 				Description: "Custom string values to be merged with the values.",
+				Deprecated: "This argument is deprecated and will be removed in the next major" +
+					" version. Use `set` argument with `type` equals to `string`",
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"name": {
@@ -324,6 +340,7 @@ func resourceRelease() *schema.Resource {
 			"metadata": {
 				Type:        schema.TypeList,
 				Computed:    true,
+				MaxItems:    1,
 				Description: "Status of the deployed release.",
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
@@ -355,7 +372,7 @@ func resourceRelease() *schema.Resource {
 						"values": {
 							Type:        schema.TypeString,
 							Computed:    true,
-							Description: "The raw yaml values used for the chart.",
+							Description: "Set of extra values, added to the chart. The sensitive data is cloaked. JSON encoded.",
 						},
 					},
 				},
@@ -649,13 +666,11 @@ func setIDAndMetadataFromRelease(d *schema.ResourceData, r *release.Release) err
 		return err
 	}
 
-	c, err := json.Marshal(r.Config)
-
+	cloakSetValues(r.Config, d)
+	values, err := json.Marshal(r.Config)
 	if err != nil {
 		return err
 	}
-
-	json := string(c)
 
 	return d.Set("metadata", []map[string]interface{}{{
 		"name":      r.Name,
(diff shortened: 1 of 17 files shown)


aareet:
Seems this also closes https://github.com/terraform-providers/terraform-provider-helm/issues/251

helm/resource_release.go
```diff
@@ -136,6 +136,14 @@ func resourceRelease() *schema.Resource {
 							Type:     schema.TypeString,
 							Required: true,
 						},
+						"type": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Default:  "auto",
```
jrhouston: When upgrading this causes a diff even though nothing has actually changed, which is confusing. I propose removing the default, as your `getValue()` func considers `"auto"` and `""` equivalent.

helm/resource_release.go
```diff
@@ -649,24 +668,53 @@ func setIDAndMetadataFromRelease(d *schema.ResourceData, r *release.Release) err
...
+}
+
+const sensitiveContentValue = "(sensitive value)"
+
+func cloakSetValue(config map[string]interface{}, key string) {
+	parts := strings.Split(key, ".")
+	m := config
+
+	var last = len(parts) - 1
+	for i, part := range parts {
```
jrhouston: I found this a bit hard to follow, what about this? 

```go
func cloakSetValue(values map[string]interface{}, valuePath string) {
	pathKeys := strings.Split(valuePath, ".")
	sensitiveKey := pathKeys[len(pathKeys)-1]
	parentPathKeys := pathKeys[:len(pathKeys)-1]

	m := values
	for _, key := range parentPathKeys {
		v, ok := m[key].(map[string]interface{})
		if !ok {
			return
		}
		m = v
	}

	m[sensitiveKey] = sensitiveContentValue
}
```

helm/resource_release.go
```diff
@@ -797,18 +842,49 @@ func getValues(d resourceGetter) (map[string]interface{}, error) {
...
-		for _, raw := range d.Get("set_sensitive").(*schema.Set).List() {
-			set := raw.(map[string]interface{})
-			yamlString = strings.Replace(yamlString, set["value"].(string), "<SENSITIVE>", -1)
+	return base, logValues(base, d)
+}
+
+func getValue(base, set map[string]interface{}) error {
+	name := set["name"].(string)
+	value := set["value"].(string)
+	typ := set["type"].(string)
```
jrhouston: What about `valueType` here instead? `typ` is a bit meaningless

helm/resource_release_test.go
```diff
@@ -714,6 +758,82 @@ func TestGetValues(t *testing.T) {
...
+		t.Fatalf("error parsing returned yaml: %s", err)
+		return
+	}
+
+	if values["foo"] != "42" {
+		t.Fatalf("error merging values, expected %q, got %s", "42", values["foo"])
+	}
+}
+
+func TestCloakSetValues(t *testing.T) {
```
jrhouston, line 783: Given that these functions are private to the package and are an implementation detail of the helm_release resource, I think it would have been fine to test this function by creating a few different incantations of the resource in the acctest above. 

I can see why you wanted to test this in isolation though, so I'm happy to leave it in.

helm/resource_release.go
```diff
@@ -797,18 +842,49 @@ func getValues(d resourceGetter) (map[string]interface{}, error) {
...
+		return fmt.Errorf("unexpected type: %s", typ)
+	}
 
-		log.Printf("---[ values.yaml ]-----------------------------------\n%s\n", yamlString)
+	return nil
+}
+
+func logValues(values map[string]interface{}, d resourceGetter) error {
+	// copy array to avoid change values by the cloak function.
+	asJSON, _ := json.Marshal(values)
```
jrhouston, line 864: Should handle this err and return it?

jrhouston (commented):
Thanks @mcuadros, this looks good just a few comments to chat about.

helm/resource_release.go
mcuadros: I can't imagine how this can fail since there is not any custom marshaller.

mcuadros:
done

helm/resource_release.go
```diff
@@ -154,13 +161,23 @@ func resourceRelease() *schema.Resource {
 							Required:  true,
 							Sensitive: true,
 						},
+						"type": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Default:  "auto",
```
jrhouston: Need to do this one too

helm/resource_release.go
jrhouston: Yeah you're right. The worst thing that could happen here is someone uses the function to try and log a map with types that couldn't be marshalled, at which point it would just log `null`

jrhouston (approved):

Issue #320, opened 2019-08-22
Feature Request: Option to Run helm template command to generate manifests locally and run kubectl apply 


anoopl:
I was looking for a feature to run `helm template ...` command then generate the manifests locally and apply with `kubectl apply -f`. This is required in the cases where we cannot use Tiller.
 I think we can do this with local exec provider. But it would be a good feature if implemented.

dominik-lekse:
I am currently working on this feature an expect a PR in the next few days. This feature has also been mentioned in #298.

anoopl:
Thanks @dominick-lekse waiting for the PR

dominik-lekse:
@anoopl PR https://github.com/terraform-providers/terraform-provider-helm/pull/323 is submitted and needs to be reviewed now.

anoopl:
Thanks @dominik-lekse. I will wait for it to be merged

nikhilbalekundargi:
PR #323 is not merged yet. Waiting for helm_template resource

mcuadros:
Closing this issue since is making reference to a version based on Helm 2, if this is still valid to the master branch please reopen it. Thanks.

Issue #298, opened 2019-06-27
helm_template data source (templating-only mode for helm)


dee-kryvenko:
Just to state obvious - people are using Helm ~wrong~ for templating only. We'll keep it out of this discussion whether it is a good or bad practice. To make the matters worse, people are using Helm for things completely outside of K8s, basically as a templating engine. Of course, Terraform has it's own templating engine, but sometimes you got to deal with what others already done and for whatever reason they picked Helm for it.

Good example is Spinnaker, they are using Helm just like that (i.e. `helm template | kubectl apply` in the nutshell).

To support this scenarios and enable TF to keep it's gluing role, I propose to introduce a new data source.

A data source `helm_template` would be similar to `helm_release` resource, it takes all that input and does `helm template` and outputs resulting documents. Optionally, split by `---` into different objects and packed as an array.

aaronmell:
Helm 3.0 removes tiller. Would this still be needed when Helm 3.0 drops?

dee-kryvenko:
@aaronmell the `helm_template` data source would still make sense. Tiller is not technically going away, it's just being incorporated with the K8s API server itself. There's still use cases when you'd want to have something in the middle after rendering a Helm chart and actually applying it, or just have it applied in your own custom way (for instance if you manage deployments with Spinnaker). Also there's a case when helm is used as a templating engine to produce something entirely else, not a k8s objects (leaving out of the scope "why").

`helm_apply` would be still needed to support a case to apply rendered result after extra-processing, but chances are that the reason people would need the after-processing would likely go away in Helm 3. Only time will tell...

On top of that, Helm 2 is not going to go away overnight, and splitting up the `helm_release` resource does not sounds like a big deal, so hence the pros and cons has to be weighted.

barryib:
Another way to support the so called **Tillerless** https://github.com/terraform-providers/terraform-provider-helm/pull/339

invidian:
>A resource helm_apply would be doing the same helm template thing and apply it directly to the cluster. Optionally, some of the Spinnaker's behaviour can be replicated - such as tracking objects in the TF state for the purpose of handling renaming/deleting resources.

I agree regarding `helm_template` resource, but I think applying should be then done by the `kubernetes` provider (even though it currently does not support creating custom resource IIRC), as applying seems out of scope for this provider.

[...]

PR #483, opened 2020-05-03
New data source `helm_template` to render chart templates locally (v3)

dominik-lekse:
This pull request adds a new data source `helm_template` to render chart templates locally. This implementation is relies on the current Helm v3 client. There has been a previous version of this pull request which was based on the Helm v2 client (refer to https://github.com/terraform-providers/terraform-provider-helm/pull/323). This pull request has been started from scratch since there have been lots of changes in `helm template` between v2 and v3.

### Description

- `helm_template` renders chart templates locally and exposes the rendered manifests in the data source attributes. `helm_template` mimics the functionality of the helm template command. 
- Arguments for `helm_template` resemble the attributes of `helm_release` when applicable and the underlying `helm template` command.
- Computed attribute `manifests` provides a map of the rendered chart templates.
- Computed attribute `manifest_bundle` provides the concatenated rendered chart templates. This corresponds to the output of the `helm template` command.
- In contrast to the v2 version of Computed attribute `notes` provides the rendered notes if the chart contains a `NOTES.txt`.

### Notes

- The attributes of the data source aim to be as close a possible to the parameters of the underlying `helm template` command.
- In the Helm implementation, `helm template` inherits all flags from `helm install` (via `addInstallFlags`). I doubt that it semantically makes sense to inherit all flags and expect that there will be a refinement in future Helm version. For the moment, all flags of `helm template` are resembled in the attributes of the data source. I think the implementation of the client must be refined before we refine this data source.

### Tasks

- [x] Implementation of `helm_template` data source
- [x] Test `TestAccDataTemplate_basic`
- [x] Test `TestAccDataTemplate_templates`
- [x] Documentation

### References

- https://github.com/terraform-providers/terraform-provider-helm/pull/323
- https://github.com/terraform-providers/terraform-provider-helm/issues/320
- https://github.com/terraform-providers/terraform-provider-helm/issues/298


dominik-lekse:
I do not think the GitHub action error

```
##[error]HttpError: Bad credentials
##[error]Bad credentials
```

is related to these changes.

Sytten:
Any chance this could land soon?

danylohetmantsev:
@aareet Any updates on that one? could it be re-tested?

danylohetmantsev:
@aareet ?

aareet:
@dominik-lekse is the use case that is being solved by this PR covered sufficiently by the new [kubernetes-alpha provider](https://registry.terraform.io/providers/hashicorp/kubernetes-alpha/latest)? If not, could you share a bit more about why not?

mcanevet:
@aareet how would you cover this using the kubernetes-alpha provider only?
I'd really like this PR to be merged so that I'm able to do:
```hcl
data "helm_template" "foo" {
  ...
}

resource "kubernetes_manifest" "foo" {
  manifest = jsondecode(data.helm_template.argocd.json)
}
```
kubernetes_alpha will be used, but we need something to generate the manifests from the helm charts.

mcanevet:
@dominik-lekse any chance to rebase this so that the maintainers could merge it?

dominik-lekse:
The main use case of the `helm_template` data source provided in this pull request is to provide the rendered manifests/resources of a Helm chart in the context of a Terraform project. 

The advantage of this data source is that it allows to automate preparing k8s manifests from Helm charts given without deploying the manifests. I.e. it provides the option to decouple preparing/rendering and deploying/applying manifests. This is especially useful when the user has deliberately decided to not adopt Terraform to apply k8s manifests or during the development of and prototyping with Helm charts.

@aareet This use case is not covered by the kubernetes-alpha provider because the kubernetes-alpha provider applies manifests to a k8s cluster. In my opinion, the data source is a valuable extension for the provider as it provides the user the choice to not cover the manifest deployment to an actual k8s cluster in their Terraform config.

> How do we continue with this PR?

This PR has exists for quite some time with several people who have shown interest. This PR has been previously rebased from an original Helm v2 implementation from mid 2019 (https://github.com/hashicorp/terraform-provider-helm/pull/323) to todays v3 implementation.

@aareet I am willing to invest the development effort to rebase and test this PR so that the feature can be merged. As a prerequisite, there should be a decision by the maintainers whether this data source is going to be merged given its functionality. Also, there should be a commitment for a place on the roadmap and a review in reasonable time. I would like to avoid a ready-to-merge PR fall behind the main branch and/or continuous rebasing with uncertain merge prospect.

@mcanevet, @danylohetmantsev and everyone else: If you are interested in this feature, please show by upvoting and/or describing your use case.

jrhouston:
@dominik-lekse We've decided to go ahead with adding this feature to the provider. 

Do you still want to rebase the branch? If not, I am happy to pick up this branch where it is and get it merge ready

dominik-lekse:
@jrhouston Thanks for the confirmation. I will rebase and test the PR and expect it ready to be merged within the next few days.

dominik-lekse:
@jrhouston The PR is rebased. I have also revised the tests which follow the approach used in the `helm_release` tests.

helm/data_template.go
```diff
@@ -0,0 +1,594 @@
...
+				Default:     defaultTemplateAttributes["include_crds"],
+				Description: "Include CRDs in the templated output",
+			},
+			"is_upgrade": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     defaultTemplateAttributes["is_upgrade"],
+				Description: "Set .Release.IsUpgrade instead of .Release.IsInstall",
+			},
+			"templates": {
```
jrhouston: I got a bit confused by where this came from – perhaps we should call it `show_only` to be consistent with the flag in the `helm template` command?

helm/data_template.go
```diff
@@ -0,0 +1,594 @@
...
+
+	client := action.NewInstall(actionConfig)
+	client.ChartPathOptions = *cpo
+	client.ClientOnly = false
+	client.DryRun = true
+	client.DisableHooks = d.Get("disable_webhooks").(bool)
+	client.Wait = d.Get("wait").(bool)
+	client.Devel = d.Get("devel").(bool)
+	client.DependencyUpdate = updateDependency
+	client.Timeout = time.Duration(d.Get("timeout").(int)) * time.Second
```
jrhouston, line 426: Does it even make sense to support fields like `timeout`, `replace`, `wait` on the data source if it just does a dry-run? I get that they're part of the install action but it seems like they wouldn't do anything useful in this case.

helm/data_template.go
```diff
@@ -0,0 +1,594 @@
...
+	}
+
+	err = isChartInstallable(c)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	updateDependency := d.Get("dependency_update").(bool)
+
+	if req := c.Metadata.Dependencies; req != nil {
```
jrhouston: This was recently refactored into it's own function, see here: 
[#28109bc](https://github.com/hashicorp/terraform-provider-helm/commit/28109bcf1f70e53c87dc869e38d7e94ef5507c18#diff-bc0e50ccdcc5046a35263e48d27d2246e25a5de107a7bcf657292fb1daaab0b2R481)

helm/data_template.go
```diff
@@ -0,0 +1,594 @@
...
+	client.DisableOpenAPIValidation = d.Get("disable_openapi_validation").(bool)
+	client.Replace = d.Get("replace").(bool)
+	client.Description = d.Get("description").(string)
+	client.CreateNamespace = d.Get("create_namespace").(bool)
+
+	// The following source has been adapted from the source of the helm template command
+	// https://github.com/helm/helm/blob/v3.5.3/cmd/helm/template.go#L67
+	client.DryRun = true
+	// Do not set fixed release name like in helm template
+	//client.ReleaseName = "RELEASE-NAME"
```
jrhouston: I don't think we need to add this commented out code. I think one `// NOTE` comment is fine to explain.

helm/data_template.go
```diff
@@ -0,0 +1,594 @@
...
+			"manifests": {
+				Type:        schema.TypeMap,
+				Optional:    true,
+				Computed:    true,
+				Description: "Map of rendered chart templates indexed by the template name.",
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+			"manifest_bundle": {
```
jrhouston: To be consistent with the `helm_release` resource I would just call this attribute `manifest`.

helm/data_template.go
```diff
@@ -0,0 +1,594 @@
...
+				}
+				manifestsToRender = append(manifestsToRender, manifestKey)
+				missing = false
+			}
+
+			if missing {
+				return diag.FromErr(fmt.Errorf("could not find template %s in chart", f))
+			}
+		}
+	} else {
```
jrhouston, line 528: I would invert this to make this the conditional, i.e `if len(templates) == 0 {}` and then not have an else at all as if the length is `0` the for loop above will do nothing.

helm/data_template.go
```diff
@@ -0,0 +1,594 @@
...
+				// manifest.Name is rendered using linux-style filepath separators on Windows as
+				// well as macOS/linux.
+				manifestPathSplit := strings.Split(manifestName, "/")
+				// manifest.Path is connected using linux-style filepath separators on Windows as
+				// well as macOS/linux
+				manifestPath := strings.Join(manifestPathSplit, "/")
+
+				// if the filepath provided matches a manifest path in the
+				// chart, render that manifest
+				if matched, _ := filepath.Match(f, manifestPath); !matched {
```
jrhouston, line 517: Something you could do here is label the outer loop and invert the logic in this conditional to use `continue <LABEL>`. That way you don't need a `missing` variable, as it will be implicit that you didn't find the template if you reach the end of the inner loop. 

edit: Ah, I see this is mostly copied from the source for the helm template command. It's just a taste thing.

helm/data_template.go
```diff
@@ -0,0 +1,594 @@
...
+					continue
+				}
+				manifestsToRender = append(manifestsToRender, manifestKey)
+				missing = false
+			}
+
+			if missing {
+				return diag.FromErr(fmt.Errorf("could not find template %s in chart", f))
```
jrhouston: There is a `diag.Errorf()` helper so you don't need do this wrapping.

jrhouston: I would also put the the template path in a `%q`

jrhouston (changes requested):
Great work here @dominik-lekse thanks for contributing. I've left a few comments mostly cosmetic stuff.

helm/data_template.go
dominik-lekse: Renamed "templates" attribute to "show_only"

helm/data_template.go
dominik-lekse: To be honest, I did not think too much about the semantics of these fields. The `helm template` command supports `timeout`, `replace`, `wait` as flags and I tried to keep the data source consistent to it.

I guess `timeout` and `wait` are relevant when using it in the `validate` mode.

helm/data_template.go
dominik-lekse: Great, changed to use `checkChartDependencies`.

helm/data_template.go
dominik-lekse: Renamed to `manifest`.

helm/data_template.go
dominik-lekse: Usually I would agree. In this case, however, I think we should stay close to the original implementation of the helm template command to reflect upstream changes more straightforward.

helm/data_template.go
dominik-lekse: Same as above and I confess: "mostly copied" ;)

When rebasing this PR to the latest helm revision, it was easier to reflect the upstream changes, because I did not refactor the original implementation.

helm/data_template.go
dominik-lekse: Done

dominik-lekse:
@jrhouston Thanks for the review. Most points are addressed, see my comments above.

jrhouston (approved):

Issue #498, opened 2020-05-14
terraform stuck at refresh state during plan


amitsehgal:
I have recently added istio charts and i already have promethius, autoscaler, external secreates, flux and fluent-d cloudwatch. So total ...~30 pods 

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

### Terraform Version and Provider Version
+ provider.aws v2.61.0
+ provider.helm v1.1.1
+ provider.kubectl (unversioned)
+ provider.kubernetes v1.11.2
+ provider.null v2.1.2
+ provider.random v2.2.1

### Provider Version
+ provider.helm v1.1.1
+ provider.kubectl (unversioned)
+ provider.kubernetes v1.11.2
+ provider.null v2.1.2
+ provider.random v2.2.1
### Affected Resource(s)

- helm_release
- helm_repository
"https://kubernetes-charts.storage.googleapis.com/"
"https://charts.fluxcd.io"

### Terraform Configuration Files
```hcl
# Copy-paste your Terraform configurations here - for large Terraform configs,
# please use a service like Dropbox and share a link to the ZIP file. For
# security, you can also encrypt the files using our GPG public key.
```

### Debug Output

2020/05/14 12:15:14 [TRACE] dag/walk: vertex "provider.helm (close)" is waiting for "module.eks_addons.helm_release.cluster_autoscaler"
2020/05/14 12:15:19 [TRACE] dag/walk: vertex "root" is waiting for "provider.helm (close)"
2020/05/14 12:15:19 [TRACE] dag/walk: vertex "provider.helm (close)" is waiting for "module.eks_addons.helm_release.cluster_autoscaler"
2020/05/14 12:15:24 [TRACE] dag/walk: vertex "root" is waiting for "provider.helm (close)"
2020/05/14 12:15:24 [TRACE] dag/walk: vertex "provider.helm (close)" is waiting for "module.eks_addons.helm_release.cluster_autoscaler"
2020/05/14 12:15:29 [TRACE] dag/walk: vertex "root" is waiting for "provider.helm (close)"
[...]

PR #505, opened 2020-05-26
Redirect klog to write to file instead of stderr to avoid blocking

pepov:
### Description

Redirects klog to file instead of stderr to avoid blocking.

### Acceptance tests

I've ran the suite and it failed with a seemingly unrelated issue, that maybe connected to:
https://github.com/terraform-providers/terraform-provider-helm/issues/494

The log output:
https://gist.github.com/pepov/1b53980c5389df3ac3451c74585070da

Update: tests passed the second time:
https://gist.github.com/pepov/8b5a172240a99b331da87b6e125bd1ca

### References
This is the original issue that this PR is trying to work around:
https://github.com/terraform-providers/terraform-provider-helm/issues/498

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,11 +1,21 @@
 package main
 
 import (
+	"flag"
+
 	"github.com/hashicorp/terraform-plugin-sdk/plugin"
 	"github.com/hashicorp/terraform-provider-helm/helm"
+	"k8s.io/klog"
 )
 
 func main() {
+	flag.Parse()
+	klogFlags := flag.NewFlagSet("klog", flag.ExitOnError)
+	klog.InitFlags(klogFlags)
+	err := klogFlags.Set("logtostderr", "false")
+	if err != nil {
+		panic(err)
+	}
 	plugin.Serve(&plugin.ServeOpts{
 		ProviderFunc: helm.Provider,
 	})


pepov:
sorry for the mentions, but could someone please help pointing me to someone who would possibly be willing to take a look at this pr and the underlying issue as **it's a blocker on EKS** @mcuadros @jrhouston 

https://github.com/terraform-providers/terraform-provider-helm/issues/498

krzysztof-miemiec:
Your PR, combined with my PR (https://github.com/hashicorp/terraform-provider-helm/pull/525) seems to work "in production" - I guess that my solution regarding concurrency issue is not the most optimal, but hey.. it works after all! 😄🎉

jtyr:
I can confirm that this PR combined together with the PR #525 solved my problem with the stuck helm provider! :heart:

jrhouston (changes requested):
Ouch, this is a nasty bug. Thanks a lot for investigating this @pepov! There is definitely a deeper problem here that I'm going to investigate - if there is a problem with klog writing to stderr then this could surface in the other kubernetes providers too. 

For now we can accept this as a bandaid to stop the provider from hanging. The branch just needs to be rebased as we just moved this repo and there's some path conflicts.

pepov:
@jrhouston rebased on master, thx for looking into this!

jrhouston (approved):

PR #512, opened 2020-05-30
Do not update helm_release state if chart download/parse error

### LINKED ISSUE #472: When the release fails for any reason do not update the state
The scenario is that a Helm release fails to deploy resources due to any error. Ie. Pod health check failure or misconfiguration of helm chart. The release updates the tf state and therefor wont redeploy on next attempt. 

### Terraform Version and Provider Version
Terraform v0.12.21
Helm v3.1.2

### Provider Version
provider.helm v1.1.1

### Affected Resource(s)
- helm_release

### Expected Behavior
If any failure occurs during deployment of the release it shouldn't mark the deployment as done

### Actual Behavior
No changes are detected between new release and previous release so terraform doesn't trigger a new deployment.

mbarrien:
### Description
Do not update helm_release state if chart fails to download (server down, bad chart path, bad chart version), or chart fails to parse/load.

This is achieved by wrapping the chart downloading and parsing inside a partial resource state, inside of which we don't actually set any fields as partially complete. This prevents saving bad chart state to the state file. Most importantly, it does not update any changed fields such as "chart" or "repository" if we fail during these steps, which is valid since during these steps no state has actually been changed on the Kubernetes cluster.

I'm not convinced this is the "right" solution (since most other resources don't have to resort to partial state updates), but it does work in my uses.

### Acceptance tests

### References
Fixes #472 (or at least part of it).
Fixes regression introduced in #449.

### DIFF
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -535,6 +535,12 @@ func resourceReleaseCreate(d *schema.ResourceData, meta interface{}) error {
 func resourceReleaseUpdate(d *schema.ResourceData, meta interface{}) error {
 	m := meta.(*Meta)
 	n := d.Get("namespace").(string)
+
+	// Enable partial resource mode to prevent "chart" from being updated if
+	// it fails to download or parse. (No actual changes to the cluster
+	// are performed until after partial resource mode is disabled.)
+	d.Partial(true)
+
 	actionConfig, err := m.GetHelmConfiguration(n)
 	if err != nil {
 		return err
@@ -589,6 +595,10 @@ func resourceReleaseUpdate(d *schema.ResourceData, meta interface{}) error {
 	if err != nil {
 		return err
 	}
+
+	// We succeeded to get the chart; the next step may actually change
+	// the deployment state. Turn off partial resource mode.
+	d.Partial(false)
 
 	name := d.Get("name").(string)
 	release, err := client.Run(name, chart, values)


mcuadros:
How this improved the feature introduced by https://github.com/terraform-providers/terraform-provider-helm/pull/464, any parse error or download error should be detected on then plan phase, so what is described here should happen.

mbarrien:
Fact of the matter is that it's NOT being caught in plan phase, otherwise I wouldn't have written this. See #472 for how to reproduce. Verified a problem in TF Helm provider release 1.2.1 (the latest)

mcuadros:
Sorry, I am lost your text claims that this handles the problem of server down, bad chart path, bad chart version, but in fact, is solve a problem with the repositories? That's why I asked because this should be covered by #464.

So will be great if you write the acceptance tests.

Thanks for the PR.

mbarrien:
Did some more investigation into @mcuadros assertion that the #464 should be enough to catch the problem in the plan phase. I found that the error I'm catching here is being thrown then ignored before it reaches any of the code in that PR. The call to getChart() at https://github.com/terraform-providers/terraform-provider-helm/blob/v1.2.1/helm/resource_release.go#L641 is properly returning an error for my case described in #472 `chart "foo/metrics-server" version "2.11.1" not found in https://kubernetes-charts.storage.googleapis.com repository`, but 2 lines down in https://github.com/terraform-providers/terraform-provider-helm/blob/v1.2.1/helm/resource_release.go#L643 the error is ignored and nil is returned.

I don't know what the right behavior should be, but this PR may not be the right answer for the bug.

mbarrien:
Please see #516 for alternate fix. I need feedback to know which approach is the better one, given that #516 breaks some tests but in deprecated code. The underlying bug is cause by a "feature" introduced in an earlier PR.

In either case, this is a proven issue that causes the Terraform state file to be in an inconsistent state and is unfixable unless you know the underlying bug; I'm preventing rolling out of Terraform Helm provider at my organization until this issue is resolved.

ksvladimir:
Hi @mbarrien and @mcuadros , I hit this same problem in my CI infrastructure, and I can confirm that this patch fixes it. It would be great to see it merged!

k0t3n:
Any updates?

PR #521, opened 2020-06-04
Update various links bots after migration

aareet:
### Description

Fixes for various files after migration. Removes the file issues.yml because I want to stop automatically adding all issues to the project so that we can triage based on 👍


dak1n1 (approved):
LGTM

PR #522, opened 2020-06-04
Replace all instances of the old org with the new one

dak1n1:
Instead of using the module name workaround from step 7 of the migration RFC
(https://docs.google.com/document/d/1PQFAhpTgUhHcZxIIWapNsa0UTPOeC5y8DO0zuV1mqCk/edit?ts=5ed8d6e0)
it's more straightforward to just rename the module.

This commit replaces all instances of `terraform-providers/terraform-provider-helm` with
`hashicorp/terraform-provider-helm`.

Also ran 'go mod tidy; go mod vendor'.


alexsomesan (approved):
LGTM

This is a strange requirement, given that there is no import of that package path in the codebase. Hmm

dak1n1:
I'm running the acceptance tests on this branch in TC, to see if we can just move forward with the module rename instead of using the workaround.

dak1n1:
Test passed. This is ready for review + merge.

aareet (approved):

Issue #552, opened 2020-07-20
Add proxy support in the provider config


bugok:
### Description

Today, the way to configure a proxy for the helm provider is to define the `http_proxy` and `https_proxy` environment variables when executing terraform. I'd like to suggest to add `http_proxy` and `https_proxy` configuration options for the helm provider.

I'm not sure if this is a helm issue or a helm-provider issue, but the usage of the `http_proxy` and `https_proxy` environment variables is less common than using the `HTTP_PROXY` and `HTTPS_PROXY` environment variables. Being able to use the capitalized versions might also be useful.

### Potential Terraform Configuration

```hcl
provider "helm" {
  kubernetes {
    config_path = "/path/to/kube_cluster.yaml"
  }
  http_proxy = myproxy:8080
  https_proxy = myproxy:8080
}```

### References

None.

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

teddylear:
I'll look into issue.

teddylear:
I created a small PR for change and ran tests locally using `make build`. I'm not sure what other testing has to be added to this PR to make sure that this change is working correctly. @bugok Please let me know if I'm missing anything on the PR so that I can add and address issues.

bugok:
@teddylear: Thanks for doing this!

Looking at the code, I see a few things:

- The code seems to be what's needed. Trying to run `helm` with `HTTPS_PROXY` and `HTTP_PROXY` seem to be working - but that's unrelated to the provider code. I can check this - by building the provider myself and using it in my terraform code, but I'm not sure how to do this. Do you know where I can read about 1. building the provider myself and 2. How to run terraform with local/dev-versions of providers?
- Documentation of this change in the README.md file is missing.
- There's a failed test in the CI: https://github.com/hashicorp/terraform-provider-helm/pull/574/checks

teddylear:
[...]

PR #574, opened 2020-09-04
Add http and https proxy to provider

teddylear:
### Description

Adding http_proxy and https_proxy 

### Acceptance tests
I have ran `make build` to test locally and added environment variables to provider_test.go as with other provider configuration. I am not sure what other testing is involved. 

Output from acceptance testing:

`make test` command has tests passing locally.

### References
This is addressing issue #552


tocak:
Did you try connecting to a remote cluster over a proxy?
I did with these changes and it didn't work.
I built the provider and put it into custom location, then configured a custom provider in terraform code so it doesn't load the lib from hashicorp. Connecting didn't work.

What would really be cool is if it could read from proxy-url cluster setting in ~/kube/config.
Kubectl and Helm CLI work good with proxy-url but not the helm provider.

PR #635, opened 2020-11-30
Upgrade engineerd/setup_kind version to 0.5.0

aareet:
### Description

Upgrades the setup_kind script version to fix the tests.

### Acceptance tests
- [ ] Have you added an acceptance test for the functionality being added?
- [x] Have you run the acceptance tests on this branch?

Output from acceptance testing:

```
$ make testacc TESTARGS='-run=TestAccXXX'

...
```

### Release Note
Release note for [CHANGELOG](https://github.com/hashicorp/terraform-provider-helm/blob/master/CHANGELOG.md):

```release-note
...
```
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

PR #637, opened 2020-12-01
Set up matrix build to run acc tests against different tf versions

jrhouston:
### Description

This PR adds a matrix build to run the acceptance tests against 0.12, 0.13, and 0.14. We could probably script further to pull the latest versions from the CDN or git tags but for now its hard coded.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -8,7 +8,7 @@ on:
     - master
 
 jobs:
-  kind:
+  unit_test:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


aareet:
@jrhouston there's a deprecation in Github Actions that's fixed in setup-kind v0.5.0 - https://github.com/engineerd/setup-kind/releases. I made a PR #635  for the upgrade, but it was blocked on the fix for the stable repo deprecation. Feel free to do it in here instead.

aareet (approved):

aareet:
Depends on #638

PR #667, opened 2021-01-20
Update v2-upgrade-guide.markdown

hmlkao:
There is no argument `repository_url` for `helm_release` resource and it was a little bit confusing.

### Description
Fix in guide.


jrhouston (approved):
Thanks for fixing this!

PR #673, opened 2021-01-26
Update provider configuration docs

jrhouston:
### Description

This PR makes the docs for configuring the provider a bit more clear.

### DIFF
--- a/website/docs/index.html.markdown
+++ b/website/docs/index.html.markdown
@@ -43,7 +43,11 @@ resource "helm_release" "nginx_ingress" {
 
 ## Authentication
 
-The provider must be explicitly configured either using the provider block or using environment variables. There are two ways to configure the Helm provider.
+When not running inside a cluster, the provider must be explicitly configured either using the provider block or using environment variables. There are three ways to configure the Helm provider:
+
+1. [Using a kubeconfig file](#file-config)
+2. [Supplying credentials](#credentials-config)
+3. [Using the in-cluster config](#in-cluster-config)
 
 ### File config
 
@@ -70,7 +74,7 @@ provider "helm" {
 }
 ```
 
-### Statically defined credentials
+### Credentials config
 
 You can also **statically** define all the credentials:
 
@@ -88,9 +92,9 @@ provider "helm" {
 }
 ```
 
-### In-cluster Configuration
+### In-cluster Config
 
-The provider is able to detect when it is running inside a cluster, so in this case you do not need to specify any configuration options in the provider block.
+The provider is able to detect when it is running inside a cluster, so in this case you do not need to specify any attributes in the provider block.
 
 
 ## Argument Reference
--- a/website/docs/index.html.markdown
+++ b/website/docs/index.html.markdown
@@ -42,7 +42,7 @@ resource "helm_release" "nginx_ingress" {
 
 ## Authentication
 
-The Helm provider can get its configuration in two ways: 
+The Helm provider can get its configuration in two ways:
 
 1. _Explicitly_ by supplying attributes to the provider block. This includes:
    * [Using a kubeconfig file](#file-config)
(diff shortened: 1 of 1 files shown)


website/docs/index.html.markdown
```diff
@@ -88,10 +93,27 @@ provider "helm" {
 }
 ```
 
-### In-cluster Configuration
+### In-cluster Config
+
+The provider is able to detect when it is running inside a cluster, so in this case you do not need to specify any attributes in the provider block.
```
redeux: if you want to connect to the local kubernetes cluster.

website/docs/index.html.markdown
```diff
@@ -70,7 +75,7 @@ provider "helm" {
 }
 ```
 
-### Statically defined credentials
+### Credentials config
 
 You can also **statically** define all the credentials:
```
redeux: I wonder if the word **statically** could be misinterpreted here, when you're able to use variables.  Should it be **explicitly** or **discretely** instead?

jrhouston: Yeah I was just thinking that – I think explicit makes more sense as being able to source these attributes from variables or outputs from other resources is not static at all

redeux:
Thanks for doing this John.  

I wonder if this could be broken down as:

- File Config
- Explicit Config
  - defined provider auth attributes
- Implicit  Config
  - Environment variables
  - In-cluster

jrhouston:
> Thanks for doing this John.
> 
> I wonder if this could be broken down as:
> 
> * File Config
> * Explicit Config
>   
>   * defined provider auth attributes
> * Implicit  Config
>   
>   * Environment variables
>   * In-cluster

Yeah totally – though I think that specifying the path to the config file is still explicit. I think noting the distinction between explicit configuration in the provider block and implicitly sourcing it from the environment is a good call.

redeux:
> > Thanks for doing this John.
> > I wonder if this could be broken down as:
> > 
> > * File Config
> > * Explicit Config
> >   
> >   * defined provider auth attributes
> > * Implicit  Config
> >   
> >   * Environment variables
> >   * In-cluster
> 
> Yeah totally – though I think that specifying the path to the config file is still explicit. I think noting the distinction between explicit configuration in the provider block and implicitly sourcing it from the environment is a good call.

Agreed, file based config is a subset of explicit config.

website/docs/index.html.markdown
```diff
@@ -39,25 +39,24 @@ resource "helm_release" "nginx_ingress" {
...
 
-1. [Using a kubeconfig file](#file-config)
-2. [Supplying credentials](#credentials-config)
-3. [Using the in-cluster config](#in-cluster-config)
-4. [Exec plugins](#exec-plugins)
 
 ### File config
 
-The easiest way is to supply a path to your kubeconfig file using the `config_path` attribute or using the `KUBE_CONFIG_PATH` environment variable.
+The easiest way is to supply a path to your kubeconfig file using the `config_path` attribute or using the `KUBE_CONFIG_PATH` environment variable. A kubeconfig file may have multiple contexts. If `config_context` is specified, the provider will use the `default` context.
```
aareet: typo - If `config_context` is *not* specified

aareet: I would also add a link to https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs#config_context

website/docs/index.html.markdown
```diff
@@ -39,25 +39,24 @@ resource "helm_release" "nginx_ingress" {
 ## Requirements
 
 - You must have a Kubernetes cluster available. We recommend version 1.14.0 or higher.
```
aareet: Should we change `recommend` to `support`?

website/docs/index.html.markdown
```diff
@@ -100,7 +97,9 @@ provider "helm" {
 
 ### In-cluster Config
 
-The provider is able to detect when it is running inside a cluster, so in this case you do not need to specify any attributes in the provider block if you want to connect to the local kubernetes cluster. If you want to connect to a different cluster than the one terraform is running inside, configure the provider as above.
+The provider use the `KUBERNETES_SERVICE_HOST` and `KUBERNETES_SERVICE_PORT` environment variables to detect when it is running inside a cluster, so in this case you do not need to specify any attributes in the provider block if you want to connect to the local kubernetes cluster.
```
aareet: The provider use -> uses

aareet (changes requested):
LGTM, minor typos

redeux (approved):

Issue #444, opened 2020-03-16
Support chart diffs


dippynark:
Currently if I apply a Helm provider chart and delete one of the created Kubernetes resources manually using `kubectl`, the provider does not notice the change on the next apply. Since this functionality does not seem to be in Helm, I can see why it isn't in the provider, but if it were to be supported it would be very powerful in terms of verifying a cluster is in the expected state. I suspect it could be implemented in the same way as [`kubectl diff`](https://kubernetes.io/blog/2019/01/14/apiserver-dry-run-and-kubectl-diff/#kubectl-diff).

Is this functionality something which would fit in this provider?

### Terraform Version
v0.12.20

### Affected Resource(s)
- helm_release

### Terraform Configuration Files

Any example would work, from the README:
```hcl
resource "helm_release" "my_database" {
    name      = "my-database"
    chart     = "stable/mariadb"

    set {
        name  = "mariadbUser"
        value = "foo"
    }

    set {
        name = "mariadbPassword"
        value = "qux"
    }

    set_string {
        name = "image.tags"
        value = "registry\\.io/terraform-provider-helm\\,example\\.io/terraform-provider-helm"
    }
}
```

### Expected Behavior
Provider should notice a change and recreate the deleted resource (maybe by upgrading the chart)

### Actual Behavior
Resource is not recreated

### Steps to Reproduce
1. `terraform apply`
2. `kubectl delete statefulset my-database-mariadb-slave`
3. `terraform apply`

aristosvo:
That would be awesome! Have you seen https://github.com/databus23/helm-diff?

jrhouston:
Thanks for opening this @dippynark. 

This is an interesting idea, but I suspect we will end up re-inventing the kubernetes provider so that we can store and render diffs for the individual kubernetes resources in the chart. 

If this is the degree of granularity you're looking for in your terraform diff, then it might be better to use the kubernetes provider and package things into terraform modules rather than using helm charts at all. Though, I will concede that at the moment this is not possible for certain use-cases like CRDs, but will be soon.

That being said, if this sort of functionality becomes part of helm's core and gets exposed through its package then we would certainly want to support it in this provider.

dippynark:
@jrhouston makes sense - have floated a PR linked above that returns a different status if resources are missing on a dry-run upgrade. This should allow a dry-run upgrade to be run when reading to determine if action is required. This won't calculate diffs but it would add quite a lot I think.

[...]

PR #702, opened 2021-03-09
Add support for storing the rendered manifest (i.e Helm diff support)

### LINKED ISSUE #79: Show the diff manifest in plan.

jrhouston:
### Description

This PR adds a feature to the provider that enables storing the rendered manifest of a release so a full diff of changes to the chart's contents can be shown at plan time. We do this by storing the rendered manifest at create time as JSON, then doing a dry run at plan time to compare.

For the moment this feature is going to be behind a flag that you have to enable either by setting the `TF_X_HELM_MANIFEST` environment variable to true, or switching it on in the provider block like this:

```hcl
provider helm {
   // ...

  experiments {
    manifest = true
  }
}
```

This is what the diff looks like:

Screen Shot 2021-03-09 at 3 48 46 PM

This feature will also pick out the Secret objects in the rendered manifest and redact the values in the `data` field, and redact any values set with `set_sensitive` from the manifest plaintext.

Closes #79 and #444

### Acceptance tests

### Release Note
Release note for [CHANGELOG](https://github.com/hashicorp/terraform-provider-helm/blob/master/CHANGELOG.md):

```release-note
Add chart diff support by storing the rendered manifest
```
### Community Note


helm/resource_release.go
```diff
@@ -689,17 +694,113 @@ func resourceDiff(ctx context.Context, d *schema.ResourceDiff, meta interface{})
...
+	// Set desired version from the Chart metadata if available
+	if len(chart.Metadata.Version) > 0 {
+		return d.SetNew("version", chart.Metadata.Version)
+	}
+
 	return d.SetNewComputed("version")
 }
 
-func setIDAndMetadataFromRelease(d *schema.ResourceData, r *release.Release) error {
+func setReleaseAttributes(d *schema.ResourceData, r *release.Release, meta interface{}) error {
```
jrhouston, line 810: I changed this because the responsibility of this function has drifted and the name had become inappropriate.

helm/manifest_json.go
```diff
@@ -0,0 +1,91 @@
...
+				secret.Data[k] = []byte(fmt.Sprintf("(sensitive value %x)", sum[:8]))
+			}
+
+			jsonbytes, err = json.Marshal(secret)
+			if err != nil {
+				return "", err
+			}
+		}
+
+		m[key] = jsonbytes
```
jrhouston, line 67: Just some context here – we have to munge the parsed manifest into a slightly different structure to capture the namespace/kind/name as a key at the top of the structure, otherwise this information gets hidden by Terraform's concise diff feature and makes it slightly confusing to read when there is many resources in the manifest.

helm/manifest_json.go
```diff
@@ -0,0 +1,99 @@
...
+		}
+
+		if gvk.Kind == "Secret" {
+			secret := corev1.Secret{}
+			err = yaml.Unmarshal([]byte(resource), &secret)
+			if err != nil {
+				return "", err
+			}
+
+			for k, v := range secret.Data {
```
jrhouston, line 56: - [ ] Need to add a unit test for this

helm/manifest_json.go
```diff
@@ -0,0 +1,99 @@
...
+
+	b, err := json.Marshal(m)
+	if err != nil {
+		return "", err
+	}
+
+	return string(b), nil
+}
+
+// hashSensitiveValue creates a hash of a sensitive value and returns the string
```
jrhouston, line 78: @alexsomesan Would appreciate your thoughts on this as an approach – it's kind of hacky but we can't lean on Terraform's sensitive feature here.

dak1n1 (approved):
Looks good! It worked in my testing, even with sensitive values.

dak1n1:
Actually, I think I found a progressive apply issue. I'll let you know the results after a little more testing. (I had one failure when using this branch along with an EKS cluster build).

helm/resource_release.go
```diff
@@ -689,17 +694,113 @@ func resourceDiff(ctx context.Context, d *schema.ResourceDiff, meta interface{})
...
+	name := d.Get("name").(string)
+	namespace := d.Get("namespace").(string)
+
+	actionConfig, err := m.GetHelmConfiguration(namespace)
+	if err != nil {
+		return err
+	}
+
+	// check if release exists
+	_, err = getRelease(m, actionConfig, name)
```
dak1n1, line 714: Unfortunately, this is calling getRelease so early in the process that it's sort of undoing the deferred provider initialization. What I mean by that is, it appears to try to contact the Kubernetes cluster before the credentials are available in the provider. And so it ends up having no provider configuration at all. [Here's my debug log](https://gist.githubusercontent.com/dak1n1/b92d6336f50a1d5bcc0afbd247895ab1/raw/084638ccaa462ea63defc687a0cfd162db01bbf6/gistfile1.txt).

These are the most relevant logs:
```
[DEBUG] getRelease for nginx-ingress-controller errored: timestamp=2021-03-20T12:47:16.506-0700
[DEBUG] Kubernetes cluster unreachable: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable: timestamp=2021-03-20T12:47:16.506-0700
```

This would mean that users would no longer be able to install their cluster (GKE, EKS, AKS, etc) in the same apply as using the helm provider. However, when I run this same test from the `main` branch, it works fine.

If you want, you can run my test here:

https://github.com/hashicorp/terraform-provider-kubernetes/tree/master/kubernetes/test-infra/eks

You'll have to use terraform version 0.14.x because the EKS module doesn't work with 0.15 yet. Also be sure to unset KUBE_CONFIG_PATH and KUBE_CONFIG_PATHS, since that overrides the explicit provider configuration.

dak1n1: It looks like it's erroring at plan time. Would it be possible to only use the new diff feature if there is an existing helm release in state, instead? That would at least allow the initial cluster build to succeed.

```
$ terraform0.14.7 plan

Error: error retrieving old release for a diff: Kubernetes cluster unreachable: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable

  on kubernetes-config/main.tf line 59, in resource "helm_release" "nginx_ingress":
  59: resource helm_release nginx_ingress {
```

jrhouston: This is a good catch, thanks for testing this @dak1n1. I've added a check so that the custom diff only runs if the release already exists – can you test again?

dak1n1: It works! 🎉

dak1n1 (approved):

helm/manifest_json.go
```diff
@@ -0,0 +1,99 @@
...
+		}
+
+		resourceMeta := resourceMeta{}
+		err = yaml.Unmarshal([]byte(resource), &resourceMeta)
+		if err != nil {
+			return "", err
+		}
+
+		gvk := resourceMeta.GetObjectKind().GroupVersionKind()
+		key := fmt.Sprintf("%s/%s", strings.ToLower(gvk.GroupKind().String()),
```
alexsomesan: I might be missing some context, but one question that comes to my mind here is "should key also include the apiVersion?". Theoretically resources of different api versions could coexist in the same release. Would that make for a key collision?

helm/manifest_json.go
alexsomesan: The only thing I would be concerned about in this setting is that partial hash collisions are a lot more likely on truncated keys. Quick googling on my side on the topic seems to support this (I'm no crypto expert).

If key length is a concern than may I suggest having a look at: https://github.com/mitchellh/hashstructure
I'm using that in the alpha provider for generating indexes in a poor man's cache and it seems to produce consistent length hashes that are still human readable. Since its recommended use-cases are for indexes, it should have good collision resistance. I'm not sure how the output length correlates to input size - that's something to look into.

alexsomesan (commented):
Good stuff!

I left some thoughts on the hashing of sensitive values and one small concern about the structure of the manifest map index.

astorath:
@alexsomesan @jrhouston Sorry to intervene, but I have some thoughts on the matter of hash functions you were discussing. 

1. `hashstructure` does not introduce anything new neither from security standpoint (it uses FNV hash by default, which is not cryptographic hash function at all) nor from code design perspective. 
2. `hashstructure` main purpose is to create hashes for complex structures, it does not have any "magic" hash algorithm. You might as well use https://golang.org/pkg/hash/fnv/#New64 and get the same result without adding any dependencies.
3. SHA-2 (SHA-256 or SHA-512) is a _good_ crypto hash by itself, you do not get any new vulnerabilities just by truncating its value besides the fact that you only need 4 billion hashes to find collision: this is the best any hash function can do with 8 bytes length.
4. Using FNV does compromise security by itself **a bit**, you can learn more here if you want: https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function (**Non-cryptographic hash** section)
5. It is not critical, cause state would already contain secret values in most cases, but FNV would allow to extract secrets from output in _some_ cases more easily.
6. If you just want get rid of that truncating part (because it looks ugly and not secure) and keep hash length of 8 bytes, I suggest using https://pkg.go.dev/golang.org/x/crypto/sha3#NewCShake128 - SHA3 SHAKE was specifically designed to be able to produce cryptographic hashes of arbitrary length. It would not give you more collision protection than SHA-2 already has, but the code would look nicer and more **secure**.

If you are concerned about security, I would suggest:
1. Store full blown HMAC SHA-256 hash in the state
2. Mark sensitive values so that terraform diff just prints them as "sensitive value" string (I don't know if it is event possible with the current API)
3. Add some random salt to the hash to counter two scenarios: 
4. (a) small sized secrets are vulnerable to any kind of hash - you must check only two values in case of boolean secret for instance
5. (b) duplicate uses of the same secret is clear in the output

PS: Just my 2 cents
PPS: All in all, I don't think current implementation is somewhat dangerous (it's essentially experimental), but before releasing it I would at least suggest to revert `hashstructure` stuff

jrhouston:
> @alexsomesan @jrhouston Sorry to intervene, but I have some thoughts on the matter of hash functions you were discussing.
> 
> 1. `hashstructure` does not introduce anything new neither from security standpoint (it uses FNV hash by default, which is not cryptographic hash function at all) nor from code design perspective.
> 2. `hashstructure` main purpose is to create hashes for complex structures, it does not have any "magic" hash algorithm. You might as well use https://golang.org/pkg/hash/fnv/#New64 and get the same result without adding any dependencies.
> 3. SHA-2 (SHA-256 or SHA-512) is a _good_ crypto hash by itself, you do not get any new vulnerabilities just by truncating its value besides the fact that you only need 4 billion hashes to find collision: this is the best any hash function can do with 8 bytes length.
> 4. Using FNV does compromise security by itself **a bit**, you can learn more here if you want: https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function (**Non-cryptographic hash** section)
> 5. It is not critical, cause state would already contain secret values in most cases, but FNV would allow to extract secrets from output in _some_ cases more easily.
> 6. If you just want get rid of that truncating part (because it looks ugly and not secure) and keep hash length of 8 bytes, I suggest using https://pkg.go.dev/golang.org/x/crypto/sha3#NewCShake128 - SHA3 SHAKE was specifically designed to be able to produce cryptographic hashes of arbitrary length. It would not give you more collision protection than SHA-2 already has, but the code would look nicer and more **secure**.
> 
> If you are concerned about security, I would suggest:
> 
> 1. Store full blown HMAC SHA-256 hash in the state
> 2. Mark sensitive values so that terraform diff just prints them as "sensitive value" string (I don't know if it is event possible with the current API)
> 3. Add some random salt to the hash to counter two scenarios:
> 4. (a) small sized secrets are vulnerable to any kind of hash - you must check only two values in case of boolean secret for instance
> 5. (b) duplicate uses of the same secret is clear in the output
> 
> PS: Just my 2 cents
> PPS: All in all, I don't think current implementation is somewhat dangerous (it's essentially experimental), but before releasing it I would at least suggest to revert `hashstructure` stuff

@astorath No apology necessary, thank you for contributing! This was a very thoughtful outline of the options here. 

I'm going to take a look at using SHA3 SHAKE instead. Although this is an experimental feature I am concerned about potentially being able to reverse the output. I'm not too concerned about hash collisions which is why I truncated the previous implementation – I just don't want to pollute the diff with a massive hash.

jrhouston:
> Mark sensitive values so that terraform diff just prints them as "sensitive value" string (I don't know if it is event possible with the current API)

Also just to follow up on this one this is currently not possible which is why we're having to solve this here. If we marked the manifest as sensitive then it would render the whole thing as `(sensitive value)` which would make this diff feature useless. 

In a future version we should be able to make this structure an object rather than a JSON string which should make it more feasible to mask sensitive values on the terraform side.

dak1n1 (approved):
The latest sensitive values hashing seems to work in my testing.

```
Terraform will perform the following actions:

  # helm_release.nginx_ingress will be updated in-place
  ~ resource "helm_release" "nginx_ingress" {
        id                         = "ingress-nginx"
      ~ manifest                   = jsonencode(
          ~ {
              ~ deployment.apps/apps/v1/ingress-nginx-controller                                                                    = {
                  ~ spec       = {
                      ~ template             = {
                          ~ spec     = {
                              ~ containers                    = [
                                  ~ {
                                      ~ args            = [
                                            # (7 unchanged elements hidden)
                                            "--validating-webhook-key=/usr/local/certificates/key",
                                          - "--maxmind-license-key=(sensitive value 842c8ce7daf5a7f9)",
                                          + "--maxmind-license-key=(sensitive value 999a5a267065e7aa)",
                                        ]
                                        # (11 unchanged elements hidden)
                                    },
                                ]
```
I ran it through the progressive apply test too. All good!

jurgen-weber-deltatre:
hey guys; 

Firstly, this is amazing. Love this feature.

Second, its super slow. Our terraform plan times have increased a fair bit. Just thought I would let you know. Please do not let that distract you from point number 1.

aareet:
@jurgen-weber-deltatre thanks, on behalf of the team :) Do you think you could run a quick benchmark with the `time` command and create a separate issue? I'm not sure how much we can do about it with the current implementation, but that would be the first step to starting to track the problem.

PR #724, opened 2021-04-09
Manifest dry-run should be behind experiment toggle

### LINKED ISSUE #719: Problem with helm when the plugin upgrade from 2.0.3 to 2.1.0 a few days ago
I had been struggled with error message similar to following for a few days:
When we ran:
> terraform plan -var-file="vars/${ENV}.tfvars"
Following error show up, and program exit:
Error: error running dry run for a diff: template: workflow-dispatcher/templates/workflow-deployment.yaml:4:38: executing "workflow/templates/workflow-deployment.yaml" at <.Values.dispatcher.name>: nil pointer evaluating interface {}.name
  on main.tf line 57, in resource "helm_release" "walle-service":
  57: resource "helm_release" "walle-service" {

This error looks like a bug introduced by upgrading the plugin from 2.0.3 to 2.1.0. On 03/30/2021, all build succeeded with downloading helm plugin 2.0.3, but on about March 31, version 2.1.0 of helm plugin is downloaded, the error above shows.

Any idea if that is a bug, or what should I do to get the variable a value?

Thank you in advance.

jrhouston:
### Description

We don't want the dry run to happen unless the `manifest` experiment is switched on. 

Fixes #719 

### Release Note

```release-note
FIX dry-run should only happen if manifest is enabled
```
### Community Note

### DIFF
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -694,100 +694,84 @@ func resourceDiff(ctx context.Context, d *schema.ResourceDiff, meta interface{})
 	}
 	debug("%s Release validated", logID)
 
-	// we don't need a custom diff if the release hasn't been created yet
-	oldStatus, _ := d.GetChange("status")
-	if oldStatus.(string) == "" {
-		return nil
-	}
-
-	// FIXME: make this a function so we can re-use it
-	// do dry-run upgrade to get diff
-	name := d.Get("name").(string)
-	namespace := d.Get("namespace").(string)
-
-	actionConfig, err := m.GetHelmConfiguration(namespace)
-	if err != nil {
-		return err
-	}
-
-	// check if release exists
-	_, err = getRelease(m, actionConfig, name)
-	if err == errReleaseNotFound {
-		if len(chart.Metadata.Version) > 0 {
-			return d.SetNew("version", chart.Metadata.Version)
+	if m.ExperimentEnabled("manifest") {
+		// we don't need a custom diff if the release hasn't been created yet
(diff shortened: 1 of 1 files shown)


dak1n1 (approved):
Looks good! Thanks for this improvement.

helm/resource_release.go
```diff
@@ -694,107 +694,93 @@ func resourceDiff(ctx context.Context, d *schema.ResourceDiff, meta interface{})
...
-
-	if m.ExperimentEnabled("manifest") {
 		jsonManifest, err := convertYAMLManifestToJSON(dry.Manifest)
 		if err != nil {
 			return err
 		}
 		manifest := redactSensitiveValues(string(jsonManifest), d)
 		d.SetNew("manifest", manifest)
 		debug("%s set manifest: %s", logID, jsonManifest)
+	} else {
```
alexsomesan, line 782: Some would say it's considered idiomatic in Go to minimise the level of indentations and the size of the sub-branches. I think if the logic is inverted in this conditional and a return added after d.Clear("manifest") it should let the rest of the code live on the same level as it was before. Just an opinion, though.

alexsomesan (approved):
Looks good. I was just exploring the idea of reducing the size of the if branches, but that's only nice-to-have.

helm/resource_release.go
jrhouston: I agree with this idiom in general, but as this is a feature toggle not business logic I feel like there is value having the code path behind the toggle be nested in an obvious way like this.

Issue #726, opened 2021-04-13
Failed deployment creation does not fail Terraform apply


citizenken:
### Terraform, Provider, Kubernetes and Helm Versions

```
Terraform version: v0.12.29
Provider version: 2.1.0
Kubernetes version: v1.18.9-eks-d1db3c
Helm version: v3.3.0
```

### Affected Resource(s)

- helm_release

### Terraform Configuration Files
```hcl
provider "helm" {
  kubernetes {
    config_path = "~/.kube/config"
  }
}

resource "helm_release" "should_fail" {
    name = "should-fail"
    chart = "./failed-deploy"
}
```
`./failed-deploy/templates/service.yaml`
```
apiVersion: v1
kind: Service
metadata:
  name: test
  namespace: doesnt-exist
spec:
  type: NodePort
  ports:
    - port: 80
      targetPort: http
      protocol: TCP
      name: http
```

### Debug Output

NOTE: In addition to Terraform debugging, please set HELM_DEBUG=1 to enable debugging info from helm.

https://gist.github.com/citizenken/9c4ee6d13a0e1f1edd7c73ee5cb53afe

### Steps to Reproduce

1. Copy module locally
2. Create a new chart with the above manifest
3. Run `terraform plan`, confirm it should create a new release
4. Run `terraform apply`, confirm that it completes successfully
5. Run `helm status should-fail`, confirm that Helm reports a failure
6. Confirm the service was not created 
7. Run `terraform plan` again, confirm that the only change is `~ status = "failed" -> "deployed"`

### Expected Behavior

I would expect that if Helm reports a deployment failed, that Terraform fails as well and cleans up the partial, failed deployment.

### Actual Behavior

Terraform reports "Complete," leaving the release in a broken state.

I think the root cause of this problem is that compared to other providers (ex. AWS), Helm allows for partial resource creation. If you have multiple resources, and only 1 fails, the others will be deployed but the status will still be "failed." From a Terraform perspective, that doesn't really work with the model other providers have that the resource is either created or it is not, there are no partial states. In this case, I think it would be better/match expectations to destroy the partial deployment and attempt to re-create on a subsequent TF run.

### References

Similar issues:
- #672
- #620

PR I've opened to fix the bug:

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
[...]

PR #727, opened 2021-04-13
Failed helm_release creation surfaces diagnostic to taint resource 

citizenken:
### Description
This PR fixes issue #726. See that issue for detailed reproduction steps.

Currently, when a Helm release is created but fails to create resources and enters a "failed" status, that failed release is recorded to TF state and no errors are returned to Terraform. Terraform then thinks the release was a success. This is an issue because the environment could be in a broken state, but Terraform will continue on with its configuration application.

This PR proposes that if a Helm release fails during the creation process, that that release should be deleted and an error returned to Terraform. No state will be persisted, so the next `terraform apply` would attempt to create that release again. This behavior is more consistent with other providers (ex. AWS) that either creates a resource or does not, they don't persist partial creations.

### Acceptance tests

### Release Note

```release-note
Failed helm_release creation surfaces diagnostic 
```
### References
 - #726

### Community Note

### DIFF
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -562,12 +562,22 @@ func resourceReleaseCreate(ctx context.Context, d *schema.ResourceData, meta int
 
 		debug("%s Release was created but returned an error, rolling back", logID)
 
-		if err != nil && exists {
-			resourceReleaseDelete(ctx, d, meta)
+		if err := setReleaseAttributes(d, rel, m); err != nil {
 			return diag.FromErr(err)
 		}
 
-		return nil
+		warnAndError := diag.Diagnostics{
+			{
+				Severity: diag.Warning,
+				Summary:  fmt.Sprintf("Helm created release \"%s\" but it is in a failed state. Persisting it for troubleshooting purposes.", client.ReleaseName),
+			},
+			{
+				Severity: diag.Error,
+				Summary:  err.Error(),
+			},
+		}
+
+		return warnAndError
 	}
 
 	err = setReleaseAttributes(d, rel, m)
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -566,10 +566,10 @@ func resourceReleaseCreate(ctx context.Context, d *schema.ResourceData, meta int
 			return diag.FromErr(err)
 		}
 
-		warnAndError := diag.Diagnostics{
+		return diag.Diagnostics{
 			{
 				Severity: diag.Warning,
-				Summary:  fmt.Sprintf("Helm created release \"%s\" but it is in a failed state. Persisting it for troubleshooting purposes.", client.ReleaseName),
+				Summary:  fmt.Sprintf("Helm release %q was created but has a failed status. Use the `helm` command to investigate the error, correct it, then run Terraform again.", client.ReleaseName),
 			},
 			{
 				Severity: diag.Error,
@@ -577,7 +577,6 @@ func resourceReleaseCreate(ctx context.Context, d *schema.ResourceData, meta int
 			},
 		}
 
-		return warnAndError
 	}
 
 	err = setReleaseAttributes(d, rel, m)
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -560,9 +560,10 @@ func resourceReleaseCreate(ctx context.Context, d *schema.ResourceData, meta int
 			return diag.FromErr(err)
 		}
 
-		debug("%s Release was created but returned an error", logID)
+		debug("%s Release was created but returned an error, rolling back", logID)
 
-		if err := setReleaseAttributes(d, rel, m); err != nil {
+		if err != nil && exists {
+			resourceReleaseDelete(ctx, d, meta)
 			return diag.FromErr(err)
 		}
 
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -560,7 +560,7 @@ func resourceReleaseCreate(ctx context.Context, d *schema.ResourceData, meta int
 			return diag.FromErr(err)
 		}
 
-		debug("%s Release was created but returned an error, rolling back", logID)
+		debug("%s Release was created but returned an error", logID)
 
 		if err := setReleaseAttributes(d, rel, m); err != nil {
 			return diag.FromErr(err)
--- a/helm/resource_release_test.go
+++ b/helm/resource_release_test.go
@@ -931,7 +931,7 @@ func testAccHelmReleaseConfigPostrender(resource, ns, name, binaryPath string) s
 
 			set {
 				name = "serviceAccount.create"
-				value = false 
+				value = false
 			}
 			set {
 				name = "service.port"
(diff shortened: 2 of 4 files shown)


citizenken:
@aareet @jrhouston I've signed the license agreement, but the status isn't reporting back to the PR. Anything I can do about it? I don't want that to block this from being reviewed.

aareet:
Hmm, I think it's because the commit doesn't appear to have an email associated with it. If you fix your gitconfig to put in an email associated with your Github account and then amend the commit, it may work

`Petti, Ken seems not to be a GitHub user.`

citizenken:
@aareet great, thanks, that worked. Had to do a force push so the tests will rerun, but should be ready for review.

helm/resource_release.go
```diff
@@ -541,9 +541,10 @@ func resourceReleaseCreate(ctx context.Context, d *schema.ResourceData, meta int
 			return diag.FromErr(err)
 		}
 
-		debug("%s Release was created but returned an error", logID)
+		debug("%s Release was created but returned an error, rolling back", logID)
 
-		if err := setReleaseAttributes(d, rel, m); err != nil {
+		if err != nil && exists {
```
jrhouston: I think this conditional might be superfluous as it's already implicit that `err != nil` because of the conditional this is nested inside, and we have a conditional above this one to handle the `!exists` case.

helm/resource_release.go
```diff
@@ -541,9 +541,10 @@ func resourceReleaseCreate(ctx context.Context, d *schema.ResourceData, meta int
...
+			resourceReleaseDelete(ctx, d, meta)
```
jrhouston: We still need to handle the diagnostics from the delete operation.

helm/resource_release.go
```diff
@@ -541,9 +541,10 @@ func resourceReleaseCreate(ctx context.Context, d *schema.ResourceData, meta int
 			return diag.FromErr(err)
 		}
 
-		debug("%s Release was created but returned an error", logID)
+		debug("%s Release was created but returned an error, rolling back", logID)
```
jrhouston: I would propose adding a warning diagnostic here to let the user know that this has happened. 

We perhaps shouldn't use the "rolling back" terminology as we're not actually doing the same thing as a `helm rollback`. Perhaps saying that the release was created in a failed state and was therefore deleted.

jrhouston (changes requested):
Thanks for contributing @citizenken. I've left a couple of comments. 

This makes sense to me in principle as it's kind of unintuitive for a Terraform resource to be "created" but in a partial or failed state but have the apply succeed. This is mostly a symptom of the helm way of doing things being slightly less deliberate than Terraform. The behaviour of the `helm install` command in this scenario is to return the error message to the user, but leave the release intact but in a failed state and leave it to the user to figure out what to do next. 

The problem I foresee with the approach in this PR is that by automatically deleting the release on the user's behalf – while we save them from additional resources being created also in a broken state down the chain – we take away their ability to debug the situation when it has gone wrong. i.e They have to go and install the chart with the `helm` command to access the broken state to investigate. 

Perhaps an alternative approach here is to leave the release intact, but return an error that forces the apply to fail?

citizenken:
@jrhouston I could back out my changes and just return the error rather than nil in this block easily enough. Maybe that would more closely model Helm’s behavior. I’ll confirm that the destroy works as expected as well.

I think this needs to be called out in a warning in the docs, specifically that leaving it in a failed state is intentional for troubleshooting.

citizenken:
@jrhouston I think this output is a good middle ground
Screen Shot 2021-04-20 at 9 30 35 AM

helm/resource_release.go
```diff
@@ -566,7 +566,18 @@ func resourceReleaseCreate(ctx context.Context, d *schema.ResourceData, meta int
 			return diag.FromErr(err)
 		}
 
-		return nil
+		warnAndError := diag.Diagnostics{
```
jrhouston: I would just put a `return` right here instead of assigning it to a variable.

helm/resource_release.go
```diff
@@ -566,7 +566,18 @@ func resourceReleaseCreate(ctx context.Context, d *schema.ResourceData, meta int
...
+			{
+				Severity: diag.Warning,
+				Summary:  fmt.Sprintf("Helm created release \"%s\" but it is in a failed state. Persisting it for troubleshooting purposes.", client.ReleaseName),
```
jrhouston: you can use `%q` instead of `\"%s\"` here

jrhouston: We could also change the wording here to be a bit more actionable – persist makes it sound like we've persisted the Terraform state of the resource, which in this case we actually haven't. In this case we probably want to tell the user that they'll need to do two things before they can run apply again:

1. Solve/fix the error
2. Delete the failed release

How about:

```
Helm release %q was created but has a failed status. Use the `helm` command to investigate the error then delete the release before running Terraform again.
```

jrhouston (changes requested):
Thanks for updating  – this seems like a sensible middle ground between rolling back automatically and allowing a failed release to satisfy a create. 

One downside of this approach is that now you have to use `helm` to delete the release before you can run apply again – `terraform destroy` wont remove it and you'll get an "already exists" error. 

I suspect this way is still the better approach because you're likely to want to poke at the broken release as your next step anyway, so automatically deleting it just creates an additional step for the user to carry out when this happens. Open to hearing other thoughts though.

helm/resource_release.go
citizenken: > One downside of this approach is that now you have to use `helm` to delete the release before you can run apply again – `terraform destroy` wont remove it and you'll get an "already exists" error.

So it looks like the release in TF gets tainted, so the next apply forces a replacement. This seems to work with my test chart. I remove the namespace definition that doesn't exist, TF deletes and re-creates the Helm release, and Helm reports the release like so:
```
NAME: should-fail
LAST DEPLOYED: Tue Apr 20 10:27:21 2021
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
```
So I think this is actually a pretty good workflow.

jrhouston: Nice! That's great then.

citizenken:
> One downside of this approach is that now you have to use `helm` to delete the release before you can run apply again – `terraform destroy` wont remove it and you'll get an "already exists" error.

So it looks like the release in TF gets tainted, so the next apply forces a replacement. This seems to work with my test chart. I remove the namespace definition that doesn't exist, TF deletes and re-creates the Helm release, and Helm reports the release like so:
```
NAME: should-fail
LAST DEPLOYED: Tue Apr 20 10:27:21 2021
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
```
So I think this is actually a pretty good workflow.

jrhouston:
Ah, I see we were still writing the state and then returning the diagnostic that's why it gets tainted. I was thinking that we were just not going to write any state at all if the release failed. You're right this workflow makes a lot more sense.

citizenken:
Exactly, which I think is good, it mirrors how Helm sees the release as failed, too, and avoids the release name collision. Thanks for taking the time to review this. What's the release schedule look like?

jrhouston:
No problem, thanks for contributing! I appreciate the back and forth. There's a couple more things we want to get in for v2.2.0 so we'll probably cut a new release later this week or next.

jrhouston (approved):

jrhouston:
Just wanted to add here that to get a release to be removed when an install fails we can set `atomic = true` on the resource and we get this error:

```bash
helm_release.test: Creating...
╷
│ Error: release test failed, and has been uninstalled due to atomic being set: namespaces "doesnt-exist" not found
│
│   on main.tf line 7:
│   (source code not available)
│
╵
```

So I think with this PR we're getting the best of both worlds as you can opt-in to the rollback if you want, and if not you'll get a partial state and a warning to let you know what to do. 👌

citizenken:
@jrhouston Ah gottcha. I wasn't sure exactly what `atomic =true` did, as the docs say `If set, installation process purges chart on fail`. I think the use of `chart` threw me off, as I was thinking in terms of releases. Maybe something like `If set, purges releases on a failed deployment` would be more clear, at least to me. Either way, I think the best of both worlds works.

PR #739, opened 2021-04-27
Use dynamic matrix for terraform versions

jrhouston:
### Description

This PR uses a dynamic matrix to get the latest versions of terraform to test against.


jrhouston:
Had to upgrade the SDK to get the tests to run because of the GPG issue in terraform-exec.

scripts/get-version-matrix.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+function get_latest_version() { 
+    curl -s https://api.github.com/repos/hashicorp/terraform/git/refs/tags | \
+        jq ".[] | .ref | split(\"/\") | .[2] | select(. | startswith(\"$1\"))" | \
+            sort -V -r | head -1 
+}
+
+echo "::set-output name=matrix::[$(get_latest_version v0.12), $(get_latest_version v0.13), $(get_latest_version v0.14), $(get_latest_version v0.15)]"
```
jrhouston, line 9: We could make this a bit more robust and turn it into it's own GitHub action for when we do this on the other repos

dak1n1 (approved):
This is a pretty clever approach! It was interesting to reference the github actions docs and check syntax, etc. It all seems to be accurate, and the bash script gives the expected output when I run it locally. I'm not sure if there's a way to actually run the action before it's merged, but LGTM.

scripts/get-version-matrix.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+function get_latest_version() { 
+    curl -s https://api.github.com/repos/hashicorp/terraform/git/refs/tags | \
```
dak1n1, line 4: I just thought of something after hitting "approve" -- sorry about the timing. But I noticed there's a releases API too. https://docs.github.com/en/rest/reference/repos#list-releases

dak1n1: Should we maybe use releases instead of tags, just in case the tagged version is not yet in the registry? They have 192 tags and 192 releases. That probably means you're safe enough to keep it as-is. But I thought I'd mention it anyway.

jrhouston:
>  I'm not sure if there's a way to actually run the action before it's merged, but LGTM.

It did run – if you look at the checks tab on this PR you can see it.

scripts/get-version-matrix.sh
jrhouston: This is a good point, although using either releases or tags is going to have the same the problem that if the binary is not yet uploaded to the releases CDN then the tests will fail because tfinstall wont be able to fetch that version yet.

Something we could do is actually hit [releases.hashicorp.com/terraform](https://releases.hashicorp.com/terraform) to get the version numbers. The problem there is that the output is HTML so we'll have to scrape it the ye old fashioned way. I know ENGSRV (see ENGSRV-055) are working on an API for release metadata but it's not yet out of the oven.

dak1n1 (approved):

Issue #737, opened 2021-04-24
helm_release metadata not cloaked when key contains "." character


camlow325:
### Terraform, Provider, Kubernetes and Helm Versions

```
Terraform version: 0.15.0, 0.14.0
Provider version: 2.1.1
Kubernetes version: 1.17.12
```

### Affected Resource(s)
- helm_release

### Terraform Configuration Files
```hcl
resource "helm_release" "example" {
  name = "redis"

  repository = "https://charts.bitnami.com/bitnami"
  chart      = "redis"

  set {
    name  = "replica.replicaCount"
    value = 0
  }

  set_sensitive {
    name  = "auth.password"
    value = "foo"
  }

  set_sensitive {
    name  = "foo\\.bar"
    value = "baz"
  }
}
```

### Steps to Reproduce
1. `terraform apply`
1. `terraform destroy`

### Expected Behavior
Apply succeeds. In the `plan` output for the destroy, the value for the `foo.bar` key under the `metadata` attribute is marked as `sensitive`:

```
  # helm_release.example will be destroyed
  - resource "helm_release" "example" {
...
      - metadata                   = [
          - {
... 
             - values      = jsonencode(
                    {
...
                      - foo.bar = "(sensitive value)"
``` 

### Actual Behavior
Apply succeeds. In the `plan` output for the destroy, the value for the `foo.bar` key under the `metadata` attribute is shown in original plain text (not marked `sensitive`):

```
  # helm_release.example will be destroyed
  - resource "helm_release" "example" {
...
      - metadata                   = [
          - {
... 
             - values      = jsonencode(
                    {
...
                      - foo.bar = "baz"
```

### Important Factoids
The value is also shown in original plain text in Terraform 0.14.x and earlier in the `plan` output ahead of an `apply` when the concise diff feature is disabled (`TF_X_CONCISE_DIFF=0`).

When the key does not have a "." character in it, (e.g., using `name = "foo.bar"` in the example above, the value is shown as `"(sensitive value)"`, as expected. In our case, though we need to set a key with a "." in its name, due to how the underlying Helm chart has been designed. We use `\\.` to escape the "." character, in order to avoid having it be misinterpreted as a delimiter for a YAML "sub-map".

It doesn't appear that using the new [sensitive](https://www.terraform.io/docs/language/functions/sensitive.html) function in Terraform 0.15 would help in this case since the sensitivity only applies to the `set_sensitive` block, not the `metadata` attribute.

### References

[...]

PR #746, opened 2021-05-15
Fix redacting sensitive values that uses backslashes #737

pviniciusfm:
### Description

Our team has found a security issue on this provider when using field names with backslashes. 
Essentially, set_sentitive doesn't work if we use backslashes because the code was splitting any dot character.

The solution was to ignore splitting when a backslash is added, and remove "\\" when trying to find the set values.

Go doesn't have a look behind feature in regex. Thus, I opted to write a string tokenizer to ignore "." that are prefixed with "\".

### Acceptance tests

### Release Note

```release-note
Fix issue #737 when redacting values that uses backslash characters in field name.
```

### References
#737 

### Community Note

### DIFF
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -875,13 +875,35 @@ func cloakSetValues(config map[string]interface{}, d resourceGetter) {
 
 const sensitiveContentValue = "(sensitive value)"
 
+func tokenizeCloakElements(valuePath string) []string {
+	var res []string
+	beginSubstr, endSubstr := 0, 0
+
+	for idx := range valuePath {
+		if (valuePath[idx] == '.') && (idx > 0) {
+			if valuePath[idx-1] != '\\' {
+				res = append(res, valuePath[beginSubstr:endSubstr])
+				beginSubstr = endSubstr + 1
+			}
+		}
+		endSubstr++
+	}
+
+	if endSubstr-beginSubstr > 0 {
+		res = append(res, valuePath[beginSubstr:])
+	}
+
+	return res
+}
+
 func cloakSetValue(values map[string]interface{}, valuePath string) {
-	pathKeys := strings.Split(valuePath, ".")
+	pathKeys := tokenizeCloakElements(valuePath)
 	sensitiveKey := pathKeys[len(pathKeys)-1]
 	parentPathKeys := pathKeys[:len(pathKeys)-1]
 
 	m := values
 	for _, key := range parentPathKeys {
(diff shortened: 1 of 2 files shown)


pviniciusfm:
Any progress on reviewing this? 
@camlow325 signed as positive for fixing the issue on their side as well.

clydetealium:
@jrhouston @dak1n1, any chance at a review?
cc: @pviniciusfm

OneMatt:
We're running into this same issue when attempting to deploy the ArgoCD helm chart with an [OIDC client secret](https://github.com/argoproj/argo-helm/blob/main/charts/argo-cd/values.yaml#L200). The key in question has a dot in it so it's not redacted properly and will show up in metadata. Any chance of getting this merged in the short term?

phyzical:
we are facing this with the exact usecase as @OneMatt 

any chance this can get nudged along? or is there any alternative that i havn't stumbled across yet?

incase its due to a lack of replication 

```
 set_sensitive {
    name  = "server.config.oidc\\.config"
    type  = "string"
    value = <<-YAML
      some_sensitive
      multi_line_value
    YAML
  }
```

above results in the following helm result
```
server:
  config:
    "oidc.config": |
        some_sensitive
        multi_line_value
```

issue is that its then exposed in the metadata output

pviniciusfm:
This issue has been like that for 2 years. I'm wondering if we still have maintainers for this project...

phyzical:
It might be that this is a bit of an edgecase as you could argue the helms that need it should just change their map structure to be friendly.

in any case my workaround for the time being is to just use until/if its fixed.

```
 lifecycle {
    ignore_changes = [metadata]
  }
```

The output does say it doesn't have any effect but my plan output begs to differ.

michelzanini:
This issue is really important. Its a security issue. 
Why is this sitting here for 3 years and no one from Hashicorp is looking into it?

Any of the top contributors can take a look here?

@arybolovlev
@jrhouston
@BBBmau

DrackThor:
Would be great if this could be addressed soon, we also run in the same issue.
so +1 here 😄

Raclaw-jl:
It would be great if someone could validate this fix

AlexMPierrat:
+1 , this is preventing us from using it :/

atmask:
This shouldn't be closed

atmask:
I've opened an updated PR to address this. If you're still facing this would love a 👍 there to try and get some traction with maintainers. TIA https://github.com/hashicorp/terraform-provider-helm/pull/1789

PR #818, opened 2022-01-11
Remove all or none of object keys in set

### LINKED ISSUE #817: Provider produces inconsistent final plan when value in set refers to unknown value that actually doesn't change after apply
```
Terraform version: 0.14.11 and 1.1.3
Provider version: 2.4.1
Kubernetes version: 1.22.4
```

### Affected Resource(s)
- helm_release

```hcl
terraform {
  required_providers {
    helm = {
      source  = "hashicorp/helm"
      version = ">= 1.2"
    }
  }
}

provider "helm" {
  kubernetes {
    config_context = "docker-desktop"
  }
}

locals {
	values = {
		"value1" = "A"
		"value2" = random_string.unknown_value.result == "AAA" ? random_string.unknown_value.result : "AAA"
	}
}

resource random_string unknown_value {
	length = 8
}

resource helm_release release {
	name	= "test-chart"
	chart	= "C:/workspace/sandbox/helm-provider/test-chart-0.1.0.tgz"
	
	dynamic set {
		for_each = local.values
		content {
			name  = set.key
			value = set.value
		}
	}
}
```

### Debug Output
N/A

### Panic Output
N/A

### Steps to Reproduce
1. `terraform apply`
2. `terraform taint random_string.unknown_value`
3. `terraform apply`

### Expected Behavior
The third apply should work out without any issues.

### Actual Behavior
The third apply fails with the error
```
Error: Provider produced inconsistent final plan

When expanding the plan for helm_release.release to include new values learned
so far during apply, provider "registry.terraform.io/hashicorp/helm" produced
an invalid new value for .set: planned set element
cty.ObjectVal(map[string]cty.Value{"name":cty.StringVal("value1"),
"type":cty.NullVal(cty.String), "value":cty.StringVal("A")}) does not
correlate with any element in actual.

This is a bug in the provider, which should be reported in the provider's own
issue tracker.
```
#### Important details
1. Dynamic set must contain at least one static value besides unknown one(s).
2. Unknown value must stay the same after apply.
3. The 'set' block must not contain 'type' attribute.

This is not an abstract case. We faced it when passing fqdn of postgres server to helm which is not changed after server gets recreated.

#### Available workarounds:
1. Define 'type' attribute explicitly with "", "auto" or "string".
2. Move blocks with unknown values to 'set_sensitive' (without any statically defined values).

#### RCA
For the third apply
1. When provider builds plan for the first time it considers 'type' attribute as removed since it is nil in Config yet "" in State.
https://github.com/hashicorp/terraform-provider-helm/blob/ac4d18cc1bebc787d060bdafdcfe3c91d2948398/vendor/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema/schema.go#L1406
So that planned change for 'value1' is:
```
name = "value1"
value = "A"
type = nil
```
2. When provider applies values learned during apply, it turns out that in fact there are no any changes and provider just gets the actual state.
https://github.com/hashicorp/terraform-provider-helm/blob/72939a9e95424fe1021342f25ef4a631a9dc47bb/vendor/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema/grpc_provider.go#L725
So that actual change for 'value1' is:
```
name = "value1"
value = "A"
type = ""
```
These changes are considered as incompatible by terraform
https://github.com/hashicorp/terraform/blob/792aa4562987e1cdbbf137a6c7d1dd1c652b8265/terraform/eval_diff.go#L87

### Important Factoids
N/A

### References
N/A

### Community Note

shakhovski:
### Description
These changes are meant to fix #817 issue. It is meant that all 'set' attributes are always present. If e.g. 'type' attribute is not defined in configuration, it however is stored in state as "". Whereas diff functions consider such an attribute as removed since it is changed from "" in state to nil in configuration. This behavior causes issues described in #817.
The idea is to consider an element is 'set' as removed only if all its attributes are marked as removed.

I'm not a go/tf developer, so my reasoning may be complete nonsense. At least it fixes #817.

### Acceptance tests

### Release Note

```release-note
...
```
### References
#817

### Community Note

### DIFF
--- a/vendor/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema/schema.go
+++ b/vendor/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema/schema.go
@@ -1335,12 +1335,21 @@ func (m schemaMap) diffSet(
 			switch t := schema.Elem.(type) {
 			case *Resource:
 				// This is a complex resource
+				var subKs []string
+				newRemoved := true
 				for k2, schema := range t.Schema {
 					subK := fmt.Sprintf("%s.%s.%s", k, code, k2)
 					err := m.diff(subK, schema, diff, d, true)
 					if err != nil {
 						return err
 					}
+					subKs = append(subKs, subK)
+					if !diff.Attributes[subK].NewRemoved {
+						newRemoved = false
+					}
+				}
+				for _, subK := range subKs {
+					diff.Attributes[subK].NewRemoved = newRemoved
 				}
 			case *Schema:
 				// Copy the schema so that we can set Computed/ForceNew from


jrhouston:
hi @shakhovski thanks for opening this, but the change you've made is actually to a dependency of this project so we can't accept it. Code in the `vendor/` directory is dependencies that are brought in from the go.mod file via the `go mod vendor` command. You will have to open a PR against the dependent project [terraform-plugin-sdk](https://github.com/hashicorp/terraform-plugin-sdk).

shakhovski:
> hi @shakhovski thanks for opening this, but the change you've made is actually to a dependency of this project so we can't accept it. Code in the `vendor/` directory is dependencies that are brought in from the go.mod file via the `go mod vendor` command. You will have to open a PR against the dependent project [terraform-plugin-sdk](https://github.com/hashicorp/terraform-plugin-sdk).

@jrhouston, thanks, got it. There are 3 newer versions of terraform-plugin-sdk, so it looks like formally I'll have to check with the latest version, although at the first glance the issue is still there.
Closing this one in favor of the possible another one on terraform-plugin-sdk.

PR #882, opened 2022-06-11
Add go-changelog

BBBmau:
### Description

Add ability to autogenerate changelog from command line by running
```
make changelog
make changelog-entry
```

### Release Note

```release-note
NONE
```
### Community Note


GNUmakefile
```diff
@@ -16,11 +16,32 @@ ifneq ($(origin TRAVIS_TAG), undefined)
...
+THIS_RELEASE?=$$(git rev-parse --abbrev-ref HEAD)
+
 default: build
 
 build: fmtcheck
 	go build -v .
 
+# expected to be invoked by make gen/changelog LAST_RELEASE=gitref THIS_RELEASE=gitref
+.PHONY: changelog-build
+changelog-build:
```
jrhouston: let's just call the target `changelog`

GNUmakefile
```diff
@@ -16,11 +16,32 @@ ifneq ($(origin TRAVIS_TAG), undefined)
...
+	@echo
+	@changelog-build -last-release $(LAST_RELEASE) \
+		-entries-dir .changelog/ \
+		-changelog-template .changelog/changelog.tmpl \
+		-note-template .changelog/note.tmpl \
+		-this-release $(THIS_RELEASE)
+
+.PHONY: changelog-entry
+changelog-entry:
+	@changelog-entry -pr $(PR) -dir .changelog/
```
jrhouston: This can just be:

```suggestion
	@changelog-entry -dir .changelog/
```

Since it will open prompts for the user to fill out, and find the PR number automatically.

GNUmakefile
```diff
@@ -16,11 +16,32 @@ ifneq ($(origin TRAVIS_TAG), undefined)
...
+LAST_RELEASE?=$$(git describe --tags $$(git rev-list --tags --max-count=1))
+THIS_RELEASE?=$$(git rev-parse --abbrev-ref HEAD)
+
 default: build
 
 build: fmtcheck
 	go build -v .
 
+# expected to be invoked by make gen/changelog LAST_RELEASE=gitref THIS_RELEASE=gitref
+.PHONY: changelog-build
```
jrhouston: There is a `.PHONY` entry at the bottom of the file where you can add the target names. 

As a side note, the whole `PHONY` thing is because Makefiles were originally used for building C files and so targets were the names of actual files – anything that wasn't a file was a "phony" target.

GNUmakefile
jrhouston: In this PR we should also add the github action that fails a check until the changelog entry is added. We can likely just copy this workflow from the waypoint repository verbatim: https://github.com/hashicorp/waypoint/blob/main/.github/workflows/changelog-checker.yml

jrhouston (changes requested):

.github/workflows/CHANGELOG_GUIDE.md
```diff
@@ -0,0 +1,111 @@
+# How To Use
```
jrhouston: This file should probably go at the top level, next to `CHANGELOG.md`

.github/workflows/changelog-checker.yml
```diff
@@ -0,0 +1,35 @@
+# This workflow checks that there is either a 'pr/no-changelog' label applied to a PR
```
jrhouston, line 1: This workflow doesn't seem to have a `name` property, we should add:

```yaml
name: Changelog Entry 
```

jrhouston (changes requested):

GNUmakefile
```diff
@@ -16,11 +16,30 @@ ifneq ($(origin TRAVIS_TAG), undefined)
...
+# any branch, and this release to just be the current branch we're on.
+LAST_RELEASE?=$$(git describe --tags $$(git rev-list --tags --max-count=1))
+THIS_RELEASE?=$$(git rev-parse --abbrev-ref HEAD)
+
 default: build
 
 build: fmtcheck
 	go build -v .
 
+# expected to be invoked by make gen/changelog LAST_RELEASE=gitref THIS_RELEASE=gitref
```
jrhouston: Oh, this comment needs to be change to say `make changelog`

go.mod
```diff
@@ -18,6 +18,8 @@ require (
 	sigs.k8s.io/yaml v1.3.0
 )
 
+require github.com/vmihailenco/msgpack v4.0.4+incompatible // indirect
```
jrhouston: not sure what happened here, we don't really need to change go.mod at all for this change so you can just remove this file from the PR for now

PR #927, opened 2022-08-05
Updates docs for 'version' & 'set'

### LINKED ISSUE #883: Setting Up Annotations (key has no value error)
## Question
Hi,  I am trying to deploy grafana and prometheus helm chart using terraform. I am facing issue while trying to setup annotations for ingress controller.
Here is the configuration I am using

```
resource "helm_release" "monitoring-stack" {
  count            = var.create_monitoring_stack ? 1 : 0
  name             = "monitor"
  repository       = "https://prometheus-community.github.io/helm-charts"
  chart            = "kube-prometheus-stack"
  version          = "35.5.1"
  namespace        = local.namespace
  create_namespace = true

  values = [
    "${file(".${path.module}/values/grafana-values.yaml")}"
  ]

  set {
    name  = "grafana.ingress.enabled"
    value = "true"
  }

  set {
    name  = "grafana.ingress.hosts"
    value = "[${var.monitor_host}]"
  }

  set {
    name  = "grafana.ingress.annotations"
    value = "alb.ingress.kubernetes.io/group.name:shared-ingress"
  }
  set {
    name  = "grafana.ingress.annotations"
    value = "alb.ingress.kubernetes.io/healthcheck-path:/health"
  }
  set {
    name  = "grafana.ingress.annotations"
    value = "alb.ingress.kubernetes.io/success-codes:200"
  }
  set {
    name  = "grafana.ingress.annotations"
    value = "alb.ingress.kubernetes.io/listen-ports:'[{\"HTTP\":80},{\"HTTPS\":443}]'"
  }
  set {
    name  = "grafana.ingress.annotations"
    value = "alb.ingress.kubernetes.io/load-balancer-attributes:deletion_protection.enabled=true"
  }
  set {
    name  = "grafana.ingress.annotations"
    value = "alb.ingress.kubernetes.io/scheme:internet-facing"
  }
  set {
    name  = "grafana.ingress.annotations"
    value = "kubernetes.io/ingress.class:alb"
  }
  set {
    name  = "grafana.ingress.annotations"
    value = "alb.ingress.kubernetes.io/certificate-arn:${var.certificate_arn}"
  }
  set {
    name  = "grafana.ingress.annotations"
    value = "alb.ingress.kubernetes.io/target-type:ip"
  }
}
```

## What is the issue?
When running the apply, it is failing with the following error
```
Error: failed parsing key "grafana.ingress.annotations" with value alb.ingress.kubernetes.io/listen-ports:'[{"HTTP":80},{"HTTPS":443}]', key "{\"HTTPS\":443}]'" has no value
```

I tried escaping with `\` and `:` itself, but nothing seems to work.

Would appreciate your help on this. Thank You.

### LINKED ISSUE #924: `version` description in `helm_release` doesn't correctly describe actual behavior per notes in #102
In the helm_release docs, `version` is described this way:

> [version](https://registry.terraform.io/providers/hashicorp/helm/latest/docs/resources/release#version) - (Optional) Specify the exact chart version to install. If this is not specified, the latest version is installed.

But the latest version isn't actually installed.  From a comment in [issue 102](https://github.com/hashicorp/terraform-provider-helm/issues/102#issuecomment-621311452_):

> This is the expected behaviour - we want to encourage the practice of being explicit about the upgrade version when upgrading an installed chart. We'll take a note to improve the documentation referenced above.

I found the actual behavior by googling and discovering that issue report, so I'm following up here to note that the docs still haven't been updated to describe it that way.

I think if the intention is that people include a `version` spec to force an update, you may as well make it a required option. Leaving it optional just causes confusion. Per the comment in the referenced issue, you're not really "encouraging" people, you're forcing them, but the docs don't make that clear. Sure, it will work at initial deployment without specifying a version, but as soon as a change is needed, a version must be specified or, per #121, the `helm_release` resource won't actually trigger a change.

Personally, though, I'd suggest it would be a much nicer behavior to leave it optional and let people choose whether they want to be explicit about it, and modify the behavior so that the latest really is installed when needed.

I think chances are high that if someone is changing their chart's version number, they expect the terraform that references it to figure out that it has changed and bump it up. That's the described default behavior now. At least in my scenario, where I have the terraform and the chart in the same folder of a repo, I have to keep 2 version strings in sync across nearly-adjacent files, and it's an error-prone update workflow. I'd much rather change it in once place and let the code be smart. DRY, right? I think your current and intended behavior makes sense when the chart is hosted elsewhere, but I don't like it when it's local. :-)

BBBmau:
### Description

Updated Docs for 'version' & 'set' to help prevent any confusion when running `helm_release` in Terraform
Fixes #924 
Fixes #883 

### Release Note

```release-note
NONE
```
### Community Note

### DIFF
--- a/website/docs/d/template.html.markdown
+++ b/website/docs/d/template.html.markdown
@@ -124,7 +124,7 @@ The following arguments are supported:
 * `repository_username` - (Optional) Username for HTTP basic authentication against the repository.
 * `repository_password` - (Optional) Password for HTTP basic authentication against the repository.
 * `devel` - (Optional) Use chart development versions, too. Equivalent to version '>0.0.0-0'. If version is set, this is ignored.
-* `version` - (Optional) Specify the exact chart version to install. If this is not specified, the latest version is installed.
+* `version` - (Optional) Specify the exact chart version to install. If this is not specified, the latest version is installed. `helm_release` will not automatically grab the latest release, version must explicitly upgraded when upgrading an installed chart.
 * `namespace` - (Optional) The namespace to install the release into. Defaults to `default`.
 * `verify` - (Optional) Verify the package before installing it. Helm uses a provenance file to verify the integrity of the chart; this must be hosted alongside the chart. For more information see the [Helm Documentation](https://helm.sh/docs/topics/provenance/). Defaults to `false`.
 * `keyring` - (Optional) Location of public keys used for verification. Used only if `verify` is true. Defaults to `/.gnupg/pubring.gpg` in the location set by `home`
--- a/website/docs/r/release.html.markdown
+++ b/website/docs/r/release.html.markdown
@@ -129,6 +129,15 @@ The `set` and `set_sensitive` blocks support:
 * `value` - (Required) value of the variable to be set.
 * `type` - (Optional) type of the variable to be set. Valid options are `auto` and `string`.
 
+Since Terraform Utilizes HCL as well as Helm using the Helm Template Language, it's necessary to  escape certain characters twice in order for it to be parsed. `name` should also be the path that leads to the desired value, where `value` is the desired value that will be set. 
+
+```hcl
+set {
+    name  = "grafana.ingress.annotations\\.alb\\.ingress\\.kubernetes\\.io/group\\.name"
+    value = "shared-ingress"
+}
+```
+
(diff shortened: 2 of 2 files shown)


arybolovlev (approved):
Nice addition! 👍🏻

PR #1176, opened 2023-07-07
resolve version conflicts with local chart

### LINKED ISSUE #1157: Version conflicts between terraform and chart

```
Terraform version: 1.3.9
Provider version: 2.10.1
Kubernetes version: 1.24
```

### Affected Resource(s)
- helm_release

```hcl
resource "helm_release" "release" {
  count      = var.enable ? 1 : 0
  name       = "release"
  chart      = "release"
  version    = "1.0.0"
  repository = "${path.module}/charts/"
  namespace  = local.namespace
  wait       = true
}

```

### Expected Behavior
Should only be applied once

### Actual Behavior
If the version stated in terraform is different from the one in Chart.yaml, the Helm provider will now plan an update on every apply. This is new behaviour since 2.10.0. If you remove the version in Terraform and rely solely on the Chart.yaml, everything is fine. In pre-2.10, the version in Chart.yaml seemed to take precedence for both planning and deployment.

### References

- GH-1097
- GH-1150

### Community Note

BBBmau:
### Description

This PR checks the `repository` value to see whether it's a path or a URL. This would prevent conflicts with the version attribute since the expected action is to have the `version` attribute match the chart version if done locally.

Having a local chart will prevent the use of the `version` attribute so that no diff is shown.

Fixes #1157

### Acceptance tests
```
 mau@mau-JKDT676NCP  ~/Dev/terraform-provider-helm   local-chart-verion ● ⍟2  make testacc TESTARGS="-run TestAccResourceRelease_LocalVersion"
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test "./helm" -v -run TestAccResourceRelease_LocalVersion -timeout 10m
2023/07/27 09:55:23 Building chart repository...
2023/07/27 09:55:23 Created repository package for "broken-chart"
2023/07/27 09:55:23 Created repository package for "crds-chart"
2023/07/27 09:55:23 Created repository package for "dependency-bar"
2023/07/27 09:55:23 Created repository package for "dependency-foo"
2023/07/27 09:55:23 Created repository package for "failed-deploy"
2023/07/27 09:55:24 Created repository package for "kube-version"
2023/07/27 09:55:24 Created repository package for "test-chart"
2023/07/27 09:55:24 Created repository package for "test-chart-v2"
2023/07/27 09:55:24 Created repository package for "umbrella-chart"
2023/07/27 09:55:24 Built chart repository index
2023/07/27 09:55:24 Test repository is listening on http://localhost:55557
=== RUN   TestAccResourceRelease_LocalVersion
--- PASS: TestAccResourceRelease_LocalVersion (9.16s)
PASS
ok      github.com/hashicorp/terraform-provider-helm/helm       10.609s
```
```
 mau@mau-JKDT676NCP  ~/Dev/terraform-provider-helm   local-chart-verion ⍟3  make testacc TESTARGS="-run TestUseChartVersion"
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test "./helm" -v -run TestUseChartVersion -timeout 10m
2023/08/03 11:54:11 Building chart repository...
2023/08/03 11:54:11 Created repository package for "broken-chart"
2023/08/03 11:54:11 Created repository package for "crds-chart"
2023/08/03 11:54:11 Created repository package for "dependency-bar"
2023/08/03 11:54:11 Created repository package for "dependency-foo"
2023/08/03 11:54:11 Created repository package for "failed-deploy"
2023/08/03 11:54:11 Created repository package for "kube-version"
2023/08/03 11:54:11 Created repository package for "test-chart"
2023/08/03 11:54:11 Created repository package for "test-chart-v2"
2023/08/03 11:54:11 Created repository package for "umbrella-chart"
2023/08/03 11:54:12 Built chart repository index
2023/08/03 11:54:12 Test repository is listening on http://localhost:56084
=== RUN   TestUseChartVersion
--- PASS: TestUseChartVersion (0.00s)
PASS
ok      github.com/hashicorp/terraform-provider-helm/helm     1.113s
```

### Release Note

```release-note:bug
`helm/resource_release.go`: Fix: version conflicts when using local chart 
```
### Community Note

### DIFF
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -1169,9 +1169,11 @@ func getChart(d resourceGetter, m *Meta, name string, cpo *action.ChartPathOptio
 
 	// Checks if chart is a URL; checks if it's a valid URL to a .tgz file of the chart
 	url, err := http.Get(name)
-	contentType := url.Header.Get("Content-Type")
-	if err == nil && contentType != "binary/octet-stream" && contentType != "application/x-gzip" {
-		return nil, "", fmt.Errorf("Not an absolute URL to the .tgz of the Chart")
+	if err == nil {
+		contentType := url.Header.Get("Content-Type")
+		if contentType != "binary/octet-stream" && contentType != "application/x-gzip" {
+			return nil, "", fmt.Errorf("Not an absolute URL to the .tgz of the Chart")
+		}
 	}
 
 	path, err := cpo.LocateChart(name, m.Settings)
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -840,8 +840,8 @@ func resourceDiff(ctx context.Context, d *schema.ResourceDiff, meta interface{})
 	if d.HasChanges(recomputeMetadataFields...) {
 		d.SetNewComputed("metadata")
 	}
-	repo := d.Get("repository").(string)
-	if !isLocalChart(d.Get("chart").(string), repo) || registry.IsOCI(repo) {
+
+	if !isLocalChart(d.Get("chart").(string), d.Get("repository").(string)) {
 		if d.HasChange("version") {
 			// only recompute metadata if the version actually changes
 			// chart versioning is not consistent and some will add
@@ -1387,7 +1387,7 @@ func chartPathOptions(d resourceGetter, m *Meta, cpo *action.ChartPathOptions) (
 	cpo.Keyring = d.Get("keyring").(string)
 	cpo.RepoURL = repositoryURL
 	cpo.Verify = d.Get("verify").(bool)
-	if !isLocalChart(chartName, cpo.RepoURL) || registry.IsOCI(repository) {
+	if !isLocalChart(chartName, cpo.RepoURL) {
 		cpo.Version = version
 	}
 	cpo.Username = d.Get("repository_username").(string)
@@ -1515,11 +1515,14 @@ func valuesKnown(d *schema.ResourceDiff) bool {
 }
 
 func isLocalChart(chart string, repo string) bool {
-	_, err := os.Stat(chart)
-	_, err1 := os.Stat(repo)
-	_, err2 := url.ParseRequestURI(chart)
-	if os.IsNotExist(err) && os.IsNotExist(err1) && err2 != nil {
-		return false
+	_, urlerr := url.ParseRequestURI(chart)
+
+	if _, err := os.Stat(chart); err == nil || urlerr == nil {
+		return true
 	}
-	return true
+	if _, err := os.Stat(repo); err == nil {
+		return true
+	}
+
+	return false
 }
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -1167,6 +1167,11 @@ func getChart(d resourceGetter, m *Meta, name string, cpo *action.ChartPathOptio
 	m.Lock()
 	defer m.Unlock()
 
+	url, err := http.Get(name)
+	if err == nil && url.Header.Get("Content-Type") != "binary/octet-stream" {
+		return nil, "", fmt.Errorf("Not an absolute URL to the .tgz of the Chart")
+	}
+
 	path, err := cpo.LocateChart(name, m.Settings)
 	if err != nil {
 		return nil, "", err
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -840,11 +840,7 @@ func resourceDiff(ctx context.Context, d *schema.ResourceDiff, meta interface{})
 	if d.HasChanges(recomputeMetadataFields...) {
 		d.SetNewComputed("metadata")
 	}
-	_, err = os.Stat(d.Get("chart").(string))
-	_, err1 := os.Stat(d.Get("repository").(string))
-	_, err2 := url.ParseRequestURI(d.Get("chart").(string))
-	if os.IsNotExist(err) && os.IsNotExist(err1) && err2 != nil {
-
+	if !isLocalChart(d.Get("chart").(string), d.Get("repository").(string)) {
 		if d.HasChange("version") {
 			// only recompute metadata if the version actually changes
 			// chart versioning is not consistent and some will add
@@ -1390,10 +1386,7 @@ func chartPathOptions(d resourceGetter, m *Meta, cpo *action.ChartPathOptions) (
 	cpo.Keyring = d.Get("keyring").(string)
 	cpo.RepoURL = repositoryURL
 	cpo.Verify = d.Get("verify").(bool)
-	_, err := os.Stat(chartName)
-	_, err1 := os.Stat(cpo.RepoURL)
-	_, err2 := url.ParseRequestURI(d.Get("chart").(string))
-	if os.IsNotExist(err) && os.IsNotExist(err1) && err2 != nil {
+	if !isLocalChart(chartName, cpo.RepoURL) {
 		cpo.Version = version
 	}
 	cpo.Username = d.Get("repository_username").(string)
@@ -1516,6 +1509,16 @@ func valuesKnown(d *schema.ResourceDiff) bool {
 		if !rawPlan.GetAttr(attr).IsWhollyKnown() {
 			return false
 		}
+	}
+	return true
+}
+
+func isLocalChart(chart string, repo string) bool {
+	_, err := os.Stat(chart)
+	_, err1 := os.Stat(repo)
+	_, err2 := url.ParseRequestURI(chart)
+	if os.IsNotExist(err) && os.IsNotExist(err1) && err2 != nil {
+		return false
 	}
 	return true
 }
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -1515,11 +1515,15 @@ func valuesKnown(d *schema.ResourceDiff) bool {
 }
 
 func useChartVersion(chart string, repo string) bool {
-	_, urlerr := url.ParseRequestURI(chart)
-
-	if _, err := os.Stat(chart); err == nil || urlerr == nil {
+	// checks if chart is a URL or OCI registry
+	if _, urlerr := url.ParseRequestURI(chart); urlerr == nil && !registry.IsOCI(chart) {
 		return true
 	}
+	// checks if chart is a local chart
+	if _, err := os.Stat(chart); err == nil {
+		return true
+	}
+	// checks if repo is a local chart
 	if _, err := os.Stat(repo); err == nil {
 		return true
 	}
--- a/helm/resource_release.go
+++ b/helm/resource_release.go
@@ -840,15 +840,19 @@ func resourceDiff(ctx context.Context, d *schema.ResourceDiff, meta interface{})
 	if d.HasChanges(recomputeMetadataFields...) {
 		d.SetNewComputed("metadata")
 	}
-	if d.HasChange("version") {
-		// only recompute metadata if the version actually changes
-		// chart versioning is not consistent and some will add
-		// a `v` prefix to the chart version after installation
-		old, new := d.GetChange("version")
-		oldVersion := strings.TrimPrefix(old.(string), "v")
-		newVersion := strings.TrimPrefix(new.(string), "v")
-		if oldVersion != newVersion {
-			d.SetNewComputed("metadata")
+	_, err = url.ParseRequestURI(d.Get("repository").(string))
+	if err == nil {
+
+		if d.HasChange("version") {
+			// only recompute metadata if the version actually changes
+			// chart versioning is not consistent and some will add
+			// a `v` prefix to the chart version after installation
+			old, new := d.GetChange("version")
+			oldVersion := strings.TrimPrefix(old.(string), "v")
+			newVersion := strings.TrimPrefix(new.(string), "v")
+			if oldVersion != newVersion {
+				d.SetNewComputed("metadata")
+			}
 		}
 	}
 
@@ -1010,7 +1014,7 @@ func resourceDiff(ctx context.Context, d *schema.ResourceDiff, meta interface{})
 		debug("%s performing dry run upgrade", logID)
 		dry, err := upgrade.Run(name, chart, values)
 		if err != nil && strings.Contains(err.Error(), "has no deployed releases") {
-			if len(chart.Metadata.Version) > 0 {
+			if len(chart.Metadata.Version) > 0 && cpo.Version != "" {
 				return d.SetNew("version", chart.Metadata.Version)
 			}
 			d.SetNewComputed("version")
@@ -1384,7 +1388,10 @@ func chartPathOptions(d resourceGetter, m *Meta, cpo *action.ChartPathOptions) (
 	cpo.Keyring = d.Get("keyring").(string)
 	cpo.RepoURL = repositoryURL
 	cpo.Verify = d.Get("verify").(bool)
-	cpo.Version = version
+	_, err := url.ParseRequestURI(cpo.RepoURL)
+	if err == nil {
+		cpo.Version = version
+	}
 	cpo.Username = d.Get("repository_username").(string)
 	cpo.Password = d.Get("repository_password").(string)
 	cpo.PassCredentialsAll = d.Get("pass_credentials").(bool)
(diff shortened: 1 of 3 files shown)


helm/resource_release.go
```diff
@@ -840,15 +840,21 @@ func resourceDiff(ctx context.Context, d *schema.ResourceDiff, meta interface{})
...
-		// chart versioning is not consistent and some will add
-		// a `v` prefix to the chart version after installation
-		old, new := d.GetChange("version")
-		oldVersion := strings.TrimPrefix(old.(string), "v")
-		newVersion := strings.TrimPrefix(new.(string), "v")
-		if oldVersion != newVersion {
-			d.SetNewComputed("metadata")
+	_, err = os.Stat(d.Get("chart").(string))
+	_, err1 := os.Stat(d.Get("repository").(string))
+	_, err2 := url.ParseRequestURI(d.Get("chart").(string))
```
jrhouston: Since this is repeated twice, you could put this logic in its own function. You could also then write a test for that function 😄 

Could maybe call the function`isLocalChart` ?

jrhouston (changes requested):

tstraley:
Does this work for cases of soft version pinning (eg. `~1.47.0`) in the `helm_release`?

BBBmau:
> Does this work for cases of soft version pinning (eg. `~1.47.0`) in the `helm_release`?

Thank you for mentioning this case @tstraley, yes it will work with soft version pinning. It will still ignore whatever `version` attribute is set assuming you are not getting a chart from a repo.

helm/resource_release.go
```diff
@@ -840,15 +840,18 @@ func resourceDiff(ctx context.Context, d *schema.ResourceDiff, meta interface{})
...
-		// only recompute metadata if the version actually changes
-		// chart versioning is not consistent and some will add
-		// a `v` prefix to the chart version after installation
-		old, new := d.GetChange("version")
-		oldVersion := strings.TrimPrefix(old.(string), "v")
-		newVersion := strings.TrimPrefix(new.(string), "v")
-		if oldVersion != newVersion {
-			d.SetNewComputed("metadata")
+	repo := d.Get("repository").(string)
+	if !isLocalChart(d.Get("chart").(string), repo) || registry.IsOCI(repo) {
```
jrhouston: Is there a way we could make the check that it's an OCI part of the isLocalChart function?

edit: you might not even need to do this, see my comment below.

helm/resource_release.go
```diff
@@ -1384,7 +1387,9 @@ func chartPathOptions(d resourceGetter, m *Meta, cpo *action.ChartPathOptions) (
 	cpo.Keyring = d.Get("keyring").(string)
 	cpo.RepoURL = repositoryURL
 	cpo.Verify = d.Get("verify").(bool)
-	cpo.Version = version
+	if !isLocalChart(chartName, cpo.RepoURL) || registry.IsOCI(repository) {
```
jrhouston: I just tried with this config:

```terraform
resource "helm_release" "nginx_ingress" {
  name       = "redis"

  chart      = "oci://registry-1.docker.io/bitnamicharts/redis"

  version = "17.14.2"

  set {
    name = "architecture"
    value = "standalone"
  }
}
```

And it still installs the latest version instead of the one I've specified in `.version`. Your check here only checks if `repository` is set to an OCI registry but you can specify a full OCI URL in `.chart`.

jrhouston (changes requested):

helm/resource_release.go
```diff
@@ -1508,3 +1512,13 @@ func valuesKnown(d *schema.ResourceDiff) bool {
 	}
 	return true
 }
+
+func isLocalChart(chart string, repo string) bool {
+	_, err := os.Stat(chart)
+	_, err1 := os.Stat(repo)
+	_, err2 := url.ParseRequestURI(chart)
+	if os.IsNotExist(err) && os.IsNotExist(err1) && err2 != nil {
```
jrhouston: Instead of having this big conditional you could invert the logic in the function to check if the `chart` or `repo` _do_ exist then return early, like this:

```go
func isLocalChart(chart string, repo string) bool {
	if _, err := os.Stat(chart); err == nil {
		return true
	}
	if _, err := os.Stat(repo); err == nil {
		return true
	}
	return false
}
```

helm/resource_release_test.go
```diff
@@ -918,6 +971,29 @@ func TestGetValuesString(t *testing.T) {
...
+
+	tests := []test{
+		{chartPath: "./testdata/charts/test-chart", repositoryURL: "", isLocalChart: true},
+		{chartPath: "", repositoryURL: "https://charts.bitnami.com/bitnami", isLocalChart: false},
+		{chartPath: "redis", repositoryURL: "https://charts.bitnami.com/bitnami", isLocalChart: false},
+	}
+
+	for i, tc := range tests {
+		if result := isLocalChart(tc.chartPath, tc.repositoryURL); result != tc.isLocalChart {
+			t.Fatalf("[%v] error in isLocalChart; expected %v, got %v", i, tc.isLocalChart, result)
```
jrhouston: I was playing with this and the error output isn't very useful, I would change it to this:

```go
t.Fatalf("[%v] error in isLocalChart; expected isLocalChart(%q, %q) == %v, got %v", i, tc.chartPath, tc.repositoryURL, tc.isLocalChart, result)

```

helm/resource_release_test.go
```diff
@@ -918,6 +971,29 @@ func TestGetValuesString(t *testing.T) {
...
+	type test struct {
+		chartPath     string
+		repositoryURL string
+		isLocalChart  bool
+	}
+
+	tests := []test{
+		{chartPath: "./testdata/charts/test-chart", repositoryURL: "", isLocalChart: true},
+		{chartPath: "", repositoryURL: "https://charts.bitnami.com/bitnami", isLocalChart: false},
+		{chartPath: "redis", repositoryURL: "https://charts.bitnami.com/bitnami", isLocalChart: false},
```
jrhouston: I would add a couple more test cases here:

```go
# when the repo is a local directory
{chartPath: "testchart", repositoryURL: "./testdata/charts", isLocalChart: true},

# when the repo is an OCI registry
{chartPath: "redis", repositoryURL: "oci://registry-1.docker.io/bitnamicharts", isLocalChart: false},

# when the chart is a URL to an OCI registry 
{chartPath: "oci://registry-1.docker.io/bitnamichartsredis", repositoryURL: "", isLocalChart: false},

File: helm/resource_release_test.go [hunk 7]
```diff
@@ -918,6 +971,29 @@ func TestGetValuesString(t *testing.T) {
...
+	}
+
+	for i, tc := range tests {
+		if result := isLocalChart(tc.chartPath, tc.repositoryURL); result != tc.isLocalChart {
+			t.Fatalf("[%v] error in isLocalChart; expected %v, got %v", i, tc.isLocalChart, result)
+		}
+	}
+
+	return
```
jrhouston: Don't need the return statement here.

helm/resource_release.go
BBBmau: thank you for this. This covers cases for local charts and OCI charts. the one case that it misses is when the chart is a direct url to a `.tgz` file.
```hcl
  name  = "redis"
  chart = "https://charts.bitnami.com/bitnami/redis-10.7.16.tgz"
}
```

adding URL parse request on `.chart` covers this case

```go
func isLocalChart(chart string, repo string) bool {
	_, urlerr := url.ParseRequestURI(chart)

	if _, err := os.Stat(chart); err == nil || urlerr == nil {
		return true
	}
	if _, err := os.Stat(repo); err == nil {
		return true
	}

	return false
}
```

helm/resource_release_test.go
```diff
@@ -918,6 +971,38 @@ func TestGetValuesString(t *testing.T) {
...
+		// when the repo is a repository URL
+		{chartPath: "", repositoryURL: "https://charts.bitnami.com/bitnami", useChartVersion: false},
+		// when chartPath is chart name and repo is repository URL
+		{chartPath: "redis", repositoryURL: "https://charts.bitnami.com/bitnami", useChartVersion: false},
+		// when the chart is a URL to an .tgz file
+		{chartPath: "https://charts.bitnami.com/bitnami/redis-10.7.16.tgz", repositoryURL: "", useChartVersion: true},
+		// when the repo is an OCI registry
+		{chartPath: "redis", repositoryURL: "oci://registry-1.docker.io/bitnamicharts", useChartVersion: false},
+		// when the chart is a URL to an OCI registry
+		{chartPath: "oci://registry-1.docker.io/bitnamicharts/redis", repositoryURL: "", useChartVersion: true},
```
jrhouston: This should be `false` — it's a path to an OCI registry, not a local file or URL to an archive.

helm/resource_release.go
```diff
@@ -1508,3 +1513,16 @@ func valuesKnown(d *schema.ResourceDiff) bool {
 	}
 	return true
 }
+
+func useChartVersion(chart string, repo string) bool {
+	_, urlerr := url.ParseRequestURI(chart)
```
jrhouston: I think you probably want to be more specific in this check as this will return true when setting the chart to a chart in a OCI registry. 

It seems like you only want this to return true if chart is a URL to a helm chart archive.

helm/resource_release.go
```diff
@@ -1508,3 +1513,16 @@ func valuesKnown(d *schema.ResourceDiff) bool {
...
+
+	if _, err := os.Stat(chart); err == nil || urlerr == nil {
```
jrhouston: Because this conditional is an `a || b`, you can simply have a separate if statement above that returns true for the `urlerr` condition. 

It's easier to understand if you formulate the logic as separate conditionals that return early.

helm/resource_release_test.go
BBBmau: It was set to `true` since we are specifically getting `oci://registry-1.docker.io/bitnamicharts/redis` chart but no version is set. So if the user wants to use a specific version they would need to set it in `.version` and if none is set it will grab the latest release.

I made some changes to the test and to `useChartVersion` since the tests were before the name change so the logic didn't make sense now when switching from `IsLocalChart` -> `useChartVersion`

BBBmau: this came due to name confusion, I've fixed now to match the logic that you were mentioning. Thanks for catching this.

helm/resource_release.go
BBBmau: this was a good catch, I covered this with the following check

```go
func useChartVersion(chart string, repo string) bool {
	// checks if chart is a URL or OCI registry
	if _, urlerr := url.ParseRequestURI(chart); urlerr == nil && !registry.IsOCI(chart) {
		return true
	}
```

so it will check if the `chart` is a URL and also will check if chart is not an OCI registry

helm/resource_release.go
```diff
@@ -1508,3 +1513,20 @@ func valuesKnown(d *schema.ResourceDiff) bool {
 	}
 	return true
 }
+
+func useChartVersion(chart string, repo string) bool {
```
alexsomesan, line 1517: It's not really clear to me how the case of the URL pointing to an archive file is being handled here. Can you elaborate on that logic a bit?

BBBmau: Sure thing, a config example would be this:
```hcl
  name  = "redis"
  chart = "https://charts.bitnami.com/bitnami/redis-10.7.16.tgz"
}
```

within `useChartVersion` we check if the chart is a URL archive or not with `url.ParseRequestURI`. It returns nil if its a working URL. When `useChartVersion` returns true,  provider will be using the chart version found directly from the `chart / repository` provided whether that's a local directory or URL archive and will ignore the `.version` field. 

Before this PR the `.version` field would get included in the state even if it doesn't match with the chart that is used directly. This would cause a lot of confusion within the state file, this solution would be to ignore the use of `.version` when it's not needed such as local charts and URL archives.

tldr; if the chart is not coming from a repository or OCI where we may need to add which version we want then the `.version` attribute will be ignored.

BBBmau: Sure thing, a config example would be this:
```hcl
  name  = "redis"
  chart = "https://charts.bitnami.com/bitnami/redis-10.7.16.tgz"
}
```

within `useChartVersion` we check if the chart is a URL archive or not with `url.ParseRequestURI`. It returns nil if its a working URL. When `useChartVersion` returns true,  provider will be using the chart version found directly from the `chart / repository` provided whether that's a local directory or URL archive and will ignore the `.version` field. 

Before this PR the `.version` field would get included in the state even if it doesn't match with the chart that is used directly. This would cause a lot of confusion within the state file, this solution would be to ignore the use of `.version` when it's not needed such as local charts and URL archives.

tldr; if the chart is not coming from a repository or OCI where we may need to add which version we want then the `.version` attribute will be ignored.

helm/resource_release.go
```diff
@@ -1508,3 +1524,21 @@ func valuesKnown(d *schema.ResourceDiff) bool {
...
+	// checks if chart is a URL or OCI registry
+
+	if url, err := http.Get(chart); err == nil && !registry.IsOCI(chart) {
+		return url.Header.Get("Content-Type") == "binary/octet-stream"
```
alexsomesan: Is this to detect the chart archive use-case? Not a bad solution, but some questions come to mind:
1) Are we 100% octet-stream only shows up when dealing with an archive and no other time?
2) Has the request already been done at this point? Is the .Headers attribute already populated?
3) Let's add a small clarifying comment at the end of this line to explain what's being done here. (e.g. "// detect if URL points to a chart archive")

BBBmau: Thank you for the comments! I seemed to have forgot to include the checks within `useChartVersion`.  I added more specific checks before the chart gets downloaded within `getChart()`

```hcl
func getChart(d resourceGetter, m *Meta, name string, cpo *action.ChartPathOptions) (*chart.Chart, string, error) {
	//Load function blows up if accessed concurrently
	m.Lock()
	defer m.Unlock()

	// Checks if chart is a URL; checks if it's a valid URL to a .tgz file of the chart
	url, err := http.Get(name)
	if err == nil {
		contentType := url.Header.Get("Content-Type")

		if contentType != "binary/octet-stream" && contentType != "application/x-gzip" && contentType != "application/x-compressed-tar" {
			return nil, "", fmt.Errorf("Not an absolute URL to the .tgz of the Chart")
		}
	}
```

1. Based on the acceptance tests, the only Content-Types that can show up when using an archive would be : `octet-stream`, `x-gzip`, `x-compressed-tar`
2. By the time we reach `useChartVersion` The headers attribute would already be populated since this would be done after downloading the chart in `getChart()`

I'm wondering if we need to still include the check within `useChartVersion` since we are doing the check of whether or not the link is a valid URL Archive link within `getChart`, it will not allow the use of a link in the `chart` attribute if its not a valid archive file. Thoughts on this? I can still include the extra check in if it makes sense to add.

alexsomesan (approved):
LGTM. Might need approval from @jrhouston because he had a "request changes" comment.

helm/resource_release.go
```diff
@@ -1163,6 +1167,16 @@ func getChart(d resourceGetter, m *Meta, name string, cpo *action.ChartPathOptio
 	m.Lock()
 	defer m.Unlock()
 
+	// Checks if chart is a URL; checks if it's a valid URL to a .tgz file of the chart
```
jrhouston: I don't think this check is necessary – the LocateChart() call below will produce a similar error if you specify a URL to an invalid chart anyway.

jrhouston (changes requested):

helm/resource_release.go
BBBmau: when running it without this test, a 403 or 404 error would be outputted if the URL chart was not a valid tar file.
```
│ Error: could not download chart: failed to fetch https://charts.bitnami.com/bitnami/redis : 403 Forbidden
```

This check is mostly just to insure that what's being provided is in fact a valid chart tar URL from the user. It seemed like the 403/404 Error was a bit vague on what exactly was preventing the chart to be used.

jrhouston: I would prefer to just defer to the helm module on what error to produce if you give it a URL that isn't a helm chart. The logic below isn't very precise. For example, I can still specify a URL that will return a different kind of binary data, or an archive that isn't a helm chart at all. It also generates an additional HTTP call that has to happen at plan time.

Generating a better error message when a user specifies an invalid URL is something we can figure out in a separate PR – I don't think this is needed to solve the original `.version` issue.

jrhouston (approved):

PR #1183, opened 2023-07-12
Add "resources" to helm_release attribute

chaaz:
Spitballing a "resources" attribute for helm_release data, which will contain roughly the same data as doing a `helm status --show-resources` call. This allows us to do things like find a service hostname using e.g. `jsondecode(helm_release.mychart.resources)["v1/Service"][0].status.loadBalancer.ingress[0].hostname` or similar tricks that require actual resource lookups.


BBBmau:
Hello, thank you for opening this PR @chaaz! When it comes to adding this it would make more sense to add this into the data source of `helm_resource` rather than it's resource.

BBBmau:
> Hello, thank you for opening this PR @chaaz! When it comes to adding this it would make more sense to add this into the data source of `helm_resource` rather than it's resource.

Closing this since the implementation of this would be more suitable with what was previously discussed above. Feel free to open a PR that applies that method here.

Issue #425, opened 2020-02-27
Error: cannot re-use a name that is still in use


hazcod:
### Terraform Version
v0.12.21

### Affected Resource(s)
- helm_release

### Terraform Configuration Files
https://github.com/ironPeakServices/infrastructure/tree/master/modules/helm

### Debug Output
https://github.com/ironPeakServices/infrastructure/runs/471718289?check_suite_focus=true

### Expected Behavior
successful terraform apply

### Actual Behavior
```hcl
Error: cannot re-use a name that is still in use

  on modules/helm/istio.tf line 9, in resource "helm_release" "istio_init":
   9: resource "helm_release" "istio_init" {
```

### Steps to Reproduce
1. `terraform apply` in CI/CD of https://github.com/ironPeakServices/infrastructure

### Important Factoids
n/a

### References
https://github.com/terraform-providers/terraform-provider-helm/blob/cfda75d3bd4770fa49a2ef057fe89eb5b8d8eb69/vendor/helm.sh/helm/v3/pkg/action/install.go#L358

marpada:
I bumped into the error after cancelling a previous run (the cloud provider stack on fulfilling a PVC request). That left some orphan resources (svc,cm,sa,...), after deleting them I could re-run the apply just fine.

aqabawe:
@hazcod did you manage to solve this? I'm running into the same issue

hazcod:
@aqabawe not really :-(

dniel:
I had the same problem, and after exploring all kinds of resources in k8s, I found a secret "sh.helm.release.v1.spa-demo.v1" that had been left behind after a partial deployment of my app "spa-demo" with a crashing container. 
I had to examine every resource I could think of to look for left behind resources from the failed deploy and  remove them manually. Found most of them quickly, but the secret was easy to miss and also the service account.

pcfleischer:
workaround is definitely as stated in previous comment:

search and delete all related helm resource by name: secrets, services, deployments, etc.

marcellodesales:
Same here... I started with `helm list` and then tried to find objects... Remember to use the appropriate namespace... 

* Deleted deployment, found secret

```
$ helm list -n kube-system
NAME        	NAMESPACE  	REVISION	UPDATED                             	STATUS  	CHART             	APP VERSION

$ kubectl get deployments -n kube-system
NAME           READY   UP-TO-DATE   AVAILABLE   AGE
coredns        2/2     2            2           92m
[...]

PR #1247, opened 2023-09-07
Enable "helm upgrade --install" equivalent behavior

n-oden:
### Description

This PR adds a (potentially) idempotent "upgrade mode" to the provider, mimicking the behavior of `helm upgrade --install` as defined in https://github.com/helm/helm/blob/main/cmd/helm/upgrade.go

To wit: an `upgrade` block is added to the resource attributes, consisting of two boolean values: `enable` and `install`.

If `enabled` is true, this causes the provider to create a `*action.Upgrade` client, and attempts to perform an upgrade on the named chart.  If `install` is true, it will first create an `*action.History` client to determine if a release already exists; if it does not find one it creates an `*action.Install` client and attempts to install the release from scratch. If a release _is_ found, an upgrade is performed.  This allows the resource to potentially co-exist with, e.g., CI/CD systems that could potentially install the release out-of-band from terraform's viewpoint.

### Acceptance tests
- [X] Have you added an acceptance test for the functionality being added?

### Release Note

```release-note
ENHANCEMENT:
* `resource/helm_release`: add `upgrade` map attribute to enable idempotent release installation, addressing components of [GH-425](https://github.com/hashicorp/terraform-provider-helm/issues/425)
```
### References

* https://github.com/hashicorp/terraform-provider-helm/issues/425

### Community Note


n-oden:
As I noted over in the #425, I've released a test build of this for people to kick the tires on:

https://registry.terraform.io/providers/n-oden/helm/latest

n-oden:
Bump: could someone re-approve the checks? I've updated the PR to address the failing ones.

helm/resource_release.go
```diff
@@ -429,6 +434,15 @@ func resourceRelease() *schema.Resource {
 					},
 				},
 			},
+			"upgrade": {
+				Type:        schema.TypeMap,
```
alexsomesan: The pattern we usually follow is to have this kind of schema construct be a `TypeList` with a single element `Elem: &schema.Resource{...` which contains needed the sub-attributes. There are historical reasons for this approach rooted in limitations of the provider SDK.

In short, can you please refactor this block along the lines of "postrender" or "metadata" above?

n-oden: Thank you for the feedback, sorry it took me a bit to see this-- been a mad few days.  I'll get this sorted out.

bd-spl:
I tried to build and use it locally, the terraform validation fails with
```
│ Error: failed to read schema for module.deploy_promtail.helm_release.this in local.dev/testing/helm: provider local.dev/testing/helm has invalid schema for managed resource type "helm_release", which is a bug in the provider: "1 error occurred:
	* upgrade.install: must set Optional, Required or Computed
```
(Terraform v1.6.0 on linux_amd64)

helm/resource_release.go
```diff
@@ -385,6 +390,29 @@ func resourceRelease() *schema.Resource {
...
+					Schema: map[string]*schema.Schema{
+						"enable": {
+							Type:        schema.TypeBool,
+							Required:    true,
+							Default:     false,
+							Description: "If true, the provider will install the release at the specified version even if a release not controlled by the provider is present: this is equivalent to using the 'helm upgrade' CLI tool rather than 'helm install'.",
+						},
+						"install": {
+							Type:        schema.TypeBool,
+							Required:    false,
```
bd-spl: I believe the validation error that I posted is related to this place?

``upgrade.install: must set Optional, Required or Computed``

n-oden: Oops, sorry, will fix.

n-oden: Oops, sorry, will fix.

n-oden:
Updated, all tests passing locally.

n-oden:
@bd-spl gentle nudge? :)

bd-spl:
> @bd-spl gentle nudge? :)

I welcome this change, albeit have no merging powers here. Please try reaching out the owners of this repository.
I think the 'copywrite' check is blocking this effort?

n-oden:
@alexsomesan can I prevail on you to take a look here? I think this is ready to go.

n-oden:
@alexsomesan nudge? :)

n-oden:
@alexsomesan this is now rebased on top of the current main branch.  Is there any chance this could be reviewed before EOY?  I've got an internal project that's rather blocked by this and would like to avoid the temptation to use a forked version without a formal 👍 on the approach from your team.

n-oden:
Happy New Year everyone! @alexsomesan my apologies for being a bother, but would it be possible to get the test workflows approved?  The tests are passing locally and I believe absent any other feedback that this is ready for submission.

lodmfjord:
Is there any news on this?

n-oden:
@lodmfjord I'm honestly not sure what's going on here; we seem to have fallen into a bit of a hole. In the meantime if you'd like to kick the tires on this you can use https://registry.terraform.io/providers/n-oden/helm/0.0.2 which is a fork of the hashicorp helm provider with this PR cherry-picked on top.  Obviously the caveat there is that there's no guarantee that they'll ever merge this back upstream, or hashicorp might ask for incompatible state schema changes before they do, so using it in production is very much an at your own risk proposition and I probably can't personally commit to keeping a fork up to date in perpetuity if this PR never gets accepted.

n-oden:
@arybolovlev @SarahFrench @appilon @jrhouston my apologies for the spam, but I'm hoping to get someone to approve running the checks on this PR, which seems to have gotten a little lost.

alexsomesan:
We're having a look at this, but I personally have to familiarise myself with the helm upgrade mechanism itself before I can evaluate this PR. I'm doing that, in the background of other tasks.

n-oden:
@alexsomesan many thanks, and if I can answer any questions please don't hesitate to ask

n-oden:
@alexsomesan just checking in: I've rebased on the current upstream main branch and dealt with the merge conflicts. Could you approve the testing workflows?

n-oden:
@alexsomesan thanks! I'm a little baffled as to why the 0.12 test is failing, and the output gives no hint.  Can you re-run that with debug logging enabled?

n-oden:
FWIW I cobbled together a local test environment and was unable to reproduce the test failure using tf v0.12.31:

https://gist.github.com/n-oden/99afa11625eee21c49efddf6ebc896db

(It did eventually fail, down in `TestAccResourceRelease_delete_regression` but I suspect this is due to my jerry-rigged test container and not any actual code issues.)

n-oden:
After some more experimentationI'm gonna strongly assert that whatever was going on with the v0.12.31 test in the previous run was not a result of this PR.  I set up a parallel PR in my own repo to make the tests run there and you can see here that it's passing: https://github.com/odenio/terraform-provider-helm/pull/1

@alexsomesan I've rebased this against the current main/HEAD -- can you re-approve the test runs?

n-oden:
@BBBmau welcome aboard. I've rebased on top of the current main branch and updated to the new documentation strategy.  Feel free to ping me on the kubernetes slack if you have questions.

n-oden:
@BBBmau the failure in https://github.com/hashicorp/terraform-provider-helm/actions/runs/9781860920/job/27286412402?pr=1247 doesn't seem like it's related to my change? But if you think otherwise let me know and I'll go hunting.

BBBmau:
> @BBBmau the failure in https://github.com/hashicorp/terraform-provider-helm/actions/runs/9781860920/job/27286412402?pr=1247 doesn't seem like it's related to my change? But if you think otherwise let me know and I'll go hunting.

Yeah the failure isn't really related to your changes. I'll be looking into this in order to get this merged ASAP.

helm/resource_release_test.go
```diff
@@ -111,6 +111,91 @@ func TestAccResourceRelease_emptyVersion(t *testing.T) {
...
+
+// "upgrade" install wherein we pretend that someone else (e.g. a CI/CD system) has done the first install
+func TestAccResourceRelease_upgrade_with_install_warmstart(t *testing.T) {
+	name := randName("basic")
+	namespace := createRandomNamespace(t)
+	// Delete namespace automatically created by helm after checks
+	defer deleteNamespace(t, namespace)
+
+	// preinstall the first revision of our chart directly via the helm CLI
+	args := []string{"install", "-n", namespace, "--create-namespace", name, "./test-chart-1.2.3.tgz"}
```
BBBmau: I believe you were intending on having it be `--create-namespace namespace name`

Both `-n` and `--create-namespaces` flags perform the same action. from `helm install -h`:
```
  -n, --namespace string                namespace scope for this request
```

BBBmau:
Some things I've come across while reviewing this:

Used `bitnami/redis` chart for testing purposes. 

I attempted to do a clean install with `upgrade.enable = true` I ran into the error which informs me that I need to either:

1. perform the install without the upgrade block or
2. perform the install with upgrade block by also setting `upgrade.install = true`

This makes sense and follows the documentation written in this PR, I went ahead and performed an install without the upgrade block. I proceeded to include `upgrade.install = true` right after with no issues.

I also attempted the following which matches the `helm upgrade <RELEASE> <CHART>` which upgrades the chart to the latest version

1. installed `bitnami/redis` with no version set, marking the version as the latest (`19.6.1`)
2. explicitly add `version = 19.6.0`, changing the version `19.6.1` -> `19.6.0`
3. remove `version` to see whether the version is upgrade to `19.6.1`This results in no changes outputted by terraform

This is good news since this PR should **only** support `helm upgrade --install`, auto upgarding to the latest release is not something we want to support in the provider. This is noted [here](https://github.com/hashicorp/terraform-provider-helm/issues/102#issuecomment-621311452)

I then attempted the case that users run into which is 
```hcl
│ Error: cannot re-use a name that is still in use
│ 
│   with helm_release.upgrade,
│   on main.tf line 1, in resource "helm_release" "upgrade":
│    1: resource "helm_release" "upgrade" {
│ 
╵
```
adding `upgade.enable = true` resolves it

i've also tested `postrender` block with `upgrade` by using one of the post-renderer demos: https://github.com/thomastaylor312/advanced-helm-demos/tree/master/post-render
```hcl
resource "helm_release" "upgrade" {
  chart = "bitnami/redis"
  name = "upgrade-test"
#   version = "19.6.1"
  namespace = "mau"
  upgrade {
    enable = true
  }
  postrender {
    binary_path = "./kustomize"
  }
}
```
Looks to also be working the same as regular `helm install --post-render`

This PR looks to be in good standing, apart from the comment on the test the only other things is checking the docs. Any feedback on that? @jrhouston

helm/resource_release_test.go
n-oden: I'm slightly confused; `-n` and `--create-namespace` are not the same flag, and AFAIK the `--create-namespace` flag is a boolean and does not take an argument:

```
$ helm install -h | grep namespace
      --create-namespace                           create the release namespace if not present
  -n, --namespace string                namespace scope for this request (default "default")
```

BBBmau: ah yeah you're right, misread it then.

BBBmau:
This is also the reason for `Deferred Actions / acceptance_tests` workflow not passing on your PR: https://github.com/orgs/community/discussions/44322

n-oden:
@BBBmau oy gevalt.  Well, that's definitely a bit out of scope for this PR to fix. :)

BBBmau:
@n-oden recent commits should have fixed the `Deferred Actions` workflow from failing on PRs coming from forks. Can you rebase to test it out?

n-oden:
@BBBmau done! I'll need you to approve running the workflows.

n-oden:
@BBBmau The failure in the docs build is not me:

```
/home/runner/work/_temp/744b2ab9-9524-42d0-a52a-c19b314fdf87.sh: line 1: Making: command not found
```

The problem here is this line:

```
e5aa8edbbc (Sheneska Williams 2024-07-23 11:21:48 -0400 40)             echo "Documentation is not up to date. Please refer to the `Making Changes` in the Contribution Guide on how to properly update documentation."
```

Because the string being fed to `echo` there is double-quoted, the backticks around "Making Changes" are being treated as a shell escape.  cc @sheneska 

I'm updating this PR to fix that, but this should probably not have made it into the main branch in the first place.

BBBmau:
> @BBBmau The failure in the docs build is not me:
> 
> ```
> /home/runner/work/_temp/744b2ab9-9524-42d0-a52a-c19b314fdf87.sh: line 1: Making: command not found
> ```
> 
> The problem here is this line:
> 
> ```
> e5aa8edbbc (Sheneska Williams 2024-07-23 11:21:48 -0400 40)             echo "Documentation is not up to date. Please refer to the `Making Changes` in the Contribution Guide on how to properly update documentation."
> ```
> 
> Because the string being fed to `echo` there is double-quoted, the backticks around "Making Changes" are being treated as a shell escape. cc @sheneska
> 
> I'm updating this PR to fix that, but this should probably not have made it into the main branch in the first place.

thanks for looking into that but ideally we'd want the fix to be in a separate PR. can you include the fix in a new PR?

n-oden:
@BBBmau sure: https://github.com/hashicorp/terraform-provider-helm/pull/1450

helm/resource_release_test.go
```diff
@@ -925,6 +1010,64 @@ func testAccHelmReleaseConfigEmptyVersion(resource, ns, name string) string {
...
+func testAccHelmReleaseConfigWithUpgradeStrategyWarmstart(ns, name string) string {
+	return fmt.Sprintf(`
+		resource "helm_release" "test" {
+ 			name        = %q
+			namespace   = %q
+			description = "Test"
+  			chart       = "./test-chart-1.2.3.tgz"
+			version     = "0.1.0"
+
+			upgrade {
```
BBBmau: after some further thinking on how this upgrade is implemented, I wonder why we a block is necessary in this case when a simpler method would be to include this as a bool field:
`upgrade_install = true`

instead of having to include a whole block:
```
			upgrade {
				enable = true
				install = false
			}
```

This comes from wanting to simplify this a bit more.

BBBmau:
> @BBBmau sure: #1450

thanks! can you also remove the workflow changes from this PR?

helm/resource_release_test.go
n-oden: I had honestly gone back and forth on this myself, and what I eventually came down on the side of was trying to map the schema here as closely as possible to the way the helm cil logic works: `helm uprade` is a different top-level command from `helm install`, and `helm upgrade --install` has behavior notably different than `helm upgrade` without the `--install` flag.

I also didn't want to presuppose how other people would use the feature — I can't imagine a scenario where I would want to enable upgrade without install=true, but it's definitely an approach supported and documented by the helm CLI so I wanted to make the provider schema map to that.

n-oden:
@BBBmau done

helm/resource_release_test.go
BBBmau: Thanks for the clarification! Let's go ahead with applying this as a boolean field. I appreciate you thinking this through, in regards to wanting to support `helm upgrade` we won't have to worry too much about this since the version itself can be upgraded on its own by updating the `version` field. 

Only other reasons to include a standalone `helm upgrade` is:

1. have the ability to automatically perform an upgrade to the latest version which we would not want. (As stated in a previous comment:
>I also attempted the following which matches the `helm upgrade <RELEASE> <CHART>` which upgrades the chart to the latest version
>
>1. installed `bitnami/redis` with no version set, marking the version as the latest (`19.6.1`)
>2. explicitly add `version = 19.6.0`, changing the version `19.6.1` -> `19.6.0`
>3. remove `version` to see whether the version is upgrade to `19.6.1`This results in no changes outputted by >terraform
>
>This is good news since this PR should **only** support `helm upgrade --install`, auto upgarding to the latest release is not something we want to support in the provider. This is noted [here](https://github.com/hashicorp/terraform-provider-helm/issues/102#issuecomment-621311452)

2. Resolve the `Error: cannot re-use a name that is still in use`, though this should be resolved with `helm upgrade --install` (i.e having a boolean field `upgrade_install = true`

This will simplify it more and should lead to a smaller block of documentation necessary for this new field.

n-oden:
@BBBmau okay, updated to use a single boolean config value, updated the documentation to match, removed a now-redundant test.

helm/resource_release.go
```diff
@@ -603,54 +610,94 @@ func resourceReleaseCreate(ctx context.Context, d *schema.ResourceData, meta int
...
 
+		if cmd := d.Get("postrender.0.binary_path").(string); cmd != "" {
+			pr, err := postrender.NewExec(cmd)
+			if err != nil {
+				return diag.FromErr(err)
+			}
+			upgradeClient.PostRenderer = pr
+		}
+
+		debug("%s Upgrading chart", logID)
```
BBBmau, line 675: From running it manually once more I noticed the following flow:

1. Install `bitnami/redis` chart with `version = "19.4.0"`
2. Remove tfstate + remove `version = "19.4.0"`
3. `terraform init` -> `terraform apply`
4. It proceeds to perform an upgrade to the latest version. `19.6.1`

This is something that we'll want to address since this goes against what was discussed prior:
> This is good news since this PR should **only** support `helm upgrade --install`, auto upgarding to the latest release is not something we want to support in the provider. This is noted [here](https://github.com/hashicorp/terraform-provider-helm/issues/102#issuecomment-621311452)

I did test every other case and it works as expected, such as:

>1. installed `bitnami/redis` with no version set, marking the version as the latest (`19.6.1`)
>2. explicitly add `version = 19.6.0`, changing the version `19.6.1` -> `19.6.0`
>3. remove `version` to see whether the version is upgrade to `19.6.1`This results in no changes outputted by terraform

I'm requesting changes here since it seems like this would be where we'd want to check if the release exists and if it does use the version associated with the already existing release in order to prevent an accidental upgrade to the latest release. Let me know if you have any questions, we appreciate your effort in this! @n-oden

n-oden: I don't love the provider having explicitly different semantics than the default helm tooling, but c'est la guerre; I'll try to get this sorted today.

n-oden: Actually... I'm going to push back on this a little, because what you are asking for is directly in opposition to what your own documentation suggests is the expected behavior, which is that if the `version` field is absent, the latest version will be installed:

```go
"version": {
  Optional:    true,
  Computed:    true,
  Description: "Specify the exact chart version to install. If this is not specified, the latest version is installed.",
},
```

...which is in alignment with how the helm cli works: if the `--version` flag is not passed in and the chart is being installed from a repo, the latest version will be picked, and this is true for both `helm install` and `helm upgrade`.

Changing this behavior in the case where `upgrade_install=true` seems to me like it breaks the principle of least surprise, and I think it would be better to explicitly warn in the documentation that the combination of upgrade_install=true and version=null will potentially change the installed release.

*That all said* commit 0470be1cd does what you've requested and adds test coverage for it.

n-oden:
The failure in [acc_test (v0.15.5)](https://github.com/hashicorp/terraform-provider-helm/actions/runs/10325741386/job/28587973282?pr=1247#logs) does not look like it's in any way obviously relevant to this change; can you try re-running that job?

BBBmau:
> The failure in [acc_test (v0.15.5)](https://github.com/hashicorp/terraform-provider-helm/actions/runs/10325741386/job/28587973282?pr=1247#logs) does not look like it's in any way obviously relevant to this change; can you try re-running that job?

Yeah often times the tests can fail due to a network issue.

We'll want to add in retries to prevent us from needing to rerun the jobs all the time @JaylonmcShan03

helm/resource_release.go
BBBmau: The `version` description was a mistake on my part, this was actually brought up in the past where the documentation itself was updated. Unfortunately I did not update the description in the schema itself 😓 https://github.com/hashicorp/terraform-provider-helm/issues/924

In the past, users would try to create a `helm_release` resource where the release would _already_ be present which where they would get this panic:

```hcl
│ Error: cannot re-use a name that is still in use
│ 
│   with helm_release.upgrade,
│   on main.tf line 1, in resource "helm_release" "upgrade":
│    1: resource "helm_release" "upgrade" {
│ 
```
the ideal way to resolve this would be to run `terraform import` in order to get the already installed release into the tfstate, this also prevents any unexpected changes to the release itself with this error.

The `upgrade-install` would also resolve this error but we still want to promote being explicit in regards to version bumping, something that's been mentioned in the past https://github.com/hashicorp/terraform-provider-helm/issues/102#issuecomment-621311452

> This is the expected behaviour - we want to encourage the practice of being explicit about the upgrade version when upgrading an installed chart. We'll take a note to improve the documentation referenced above.

BBBmau:
@n-oden did the same manual test i did before and I see that it uses the latest version rather than the chart version already being used on the chart:

when running `TF_LOG=debug terraform apply`:
```hcl
32 [DEBUG] [getInstalledReleaseVersion: mau-test-pr] Chart mau-test-pr is installed as release 19.6.1: timestamp=2024-08-09T15:58:32.743-0700
2024-08-09T15:58:33.849-0700 [INFO]  provider.terraform-provider-helm_9.9.9_darwin_arm64: 2024/08/09 15:58:33 [DEBUG] [resourceDiff: mau-test-pr] Got chart redis version 19.6.1: timestamp=2024-08-09T15:58:33.849-0700
2024-08-09T15:58:33.850-0700 [INFO]  provider.terraform-provider-helm_9.9.9_darwin_arm64: 2024/08/09 15:58:33 [DEBUG] [resourceDiff: mau-test-pr] Release validated: timestamp=2024-08-09T15:58:33.850-0700
2024-08-09T15:58:33.850-0700 [INFO]  provider.terraform-provider-helm_9.9.9_darwin_arm64: 2024/08/09 15:58:33 [DEBUG] [resourceDiff: mau-test-pr] setting version to installed version 19.6.1: timestamp=2024-08-09T15:58:33.850-0700
```

This appears with this tfconfig where tfstate is empty as well as the release being installe with version being `19.4.0`

`tfconfig`:
```hcl
resource "helm_release" "test" {
  name = "mau-test-pr"
  chart = "bitnami/redis"
  # version = "19.4.0"
  upgrade_install = true
}
```

from `helm list`:
```hcl
└─(16:00:30)──> helm list                               ──(Fri,Aug09)─┘
NAME            NAMESPACE       REVISION        UPDATED                                 STATUS          CHART           APP VERSION
mau-test-pr     default         18              2024-08-09 15:58:07.823949 -0700 PDT    deployed        redis-19.4.0    7.2.5
```

n-oden:
@BBBmau okay that's weird -- I would think that the test case in [TestAccResourceRelease_upgrade_with_install_warmstart_no_version](https://github.com/hashicorp/terraform-provider-helm/pull/1247/files#diff-0c630859c0a0648a5a7cd178d9b8d011108f16648cd8fc6b651390cec95ebc56R190) would cover this exact scenario: we preinstall test-chart v1.2.3, apply the upgrade_install plan and then verify that v1.2.3 is still installed.

I'll try to figure out what's going on, but can you please attach the full debug log of your tf plan/apply sessions as well as the helm cli commands your ran and their output in a gist?

n-oden:
@BBBmau I don't seem to be able to reproduce this? Take a look at this gist: https://gist.github.com/n-oden/d54d278d3a2ffbc5e2fb81d6f00d3092

The upshot is:

- manually `helm install redis-test bitnami/redis --version 19.4.0`; `helm list` confirms that chart version 19.4.0 is installed at revision 1
- terraform plan/apply with upgrade_install=true and version=null
- `helm list` confirms that revision 2 has been installed but still chart vresion 19.4.0
- `terraform plan` confirms no changes at this point
- the string `19.4.1` does not appear in any of the log files

Can you double-check this on your side?

n-oden:
One thing that's probably worth explicitly noting is that this _does_ create a new revision, which means that if there are deployments or statefulsets in your release, it will create a kubernetes rollout event, with the associated pod restarts.  It's the moral equivalent of running:

```
$ helm install  --version 19.4.0 myredis bitnami/redis
$ helm upgrade --version 19.4.0 myredis bitnami/redis
```

...which if you do it locally you will observe does create rollout events for both the deploy and sts resources.

I'm going to assert strongly that this is expected behavior and I'll add a note about it to the documentation.

docs/resources/release.md
```diff
@@ -326,6 +327,50 @@ The `postrender` block supports two attributes:
...
+must manually inspect the installed release's values with the `helm get values` CLI command.
+
+*IMPORTANT*: Even if you are "upgrading" to the same version of the chart that is already present in the cluster,
+the `helm_release` resource will still show as "changed" in the terraform plan output, because there is no existing
+state for it to compare against. This also means that in the apply stage, the provider will in fact reinstall the
+chart, which means that if there are any
+[deployment](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/),
+[daemonset](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/) or
+[statefulset](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/) resources in the chart, they will
+be replaced, which will cause a rolling update of the pods.
```
BBBmau: I was attempting to push a commit to make minor docs changes but I don't have permissions on this PR. Can you apply the changes below:
```
When using the Helm CLI directly, it is possible to use `helm upgrade --install` to
_idempotently_ install a release.  For example, `helm upgrade --install mariadb charts/mariadb --verson 7.1.0`
will check to see if there is already a release called `mariadb`: if there is, ensure that it is set to version
7.1.0, and if there is not, install that version from scratch. (See the documentation for the
[helm upgrade](https://helm.sh/docs/helm/helm_upgrade) command for more details.)

 **NOTE:** The mechanics of this approach are subtly different from the defaults and you can easily produce unexpected or undesirable results if you are not careful:
using this approach in production is not necessarily recommended!

If upgrade mode is enabled by setting the `upgrade_install` attribute to `true`, the provider will first check to see
if a release with the given name already exists.  If the release does not already exist, the provider will perform a
from-scratch installation of the chart.  In this case, all resource attributes are honored.

However, if the release _does_ already exist, the provider will attempt to upgrade the release to
the state defined in the resource, using the same strategy as the [helm upgrade](https://helm.sh/docs/helm/helm_upgrade)
command.

When using `upgrade_install`, the `version` attribute is used to determine the version of the chart to install or
upgrade to.  If the `version` attribute is not set, the provider will attempt to determine the version of the chart
from the existing release and will use that version for the upgrade: this is to ensure that using `upgrade_install`
does not inadvertently change the version of the chart being used.

**CRITICAL**: The user-supplied values passed to the chart in the new revision will be the ones specified in the
`helm_release` resource, not the values used in the original installation of the chart. This means that if
you are using `upgrade_install` to manage a release that was originally installed with a different set of values,
you must ensure that the values in the `helm_release` resource are correct, or you may inadvertently change the
configuration of the release. Additionally, since there is no existing terraform state to compare against, you
must manually inspect the installed release's values with the `helm get values` CLI command.

**IMPORTANT**: Even if you are "upgrading" to the same version of the chart that is already present in the cluster,
the `helm_release` resource will still show as "changed" in the terraform plan output, because there is no existing
state for it to compare against. This also means that in the apply stage, the provider will in fact reinstall the
chart, which means that if there are any
[deployment](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/),
[daemonset](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/) or
[statefulset](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/) resources in the chart, they will
be replaced, which will cause a rolling update of the pods.
```

templates/resources/release.md.tmpl
```diff
@@ -72,6 +72,50 @@ The `postrender` block supports two attributes:
 * `binary_path` - (Required) relative or full path to command binary.
 * `args` - (Optional) a list of arguments to supply to the post-renderer.
 
+## Upgrade Mode Notes
+
+When using the Helm CLI directly, it is possible (and fairly common) to use `helm upgrade --install` to
```
BBBmau: same should be applied here

docs/index.md
```diff
@@ -186,10 +186,10 @@ The `kubernetes` block supports:
 * `config_context` - (Optional) Context to choose from the config file. Can be sourced from `KUBE_CTX`.
 * `proxy_url` - (Optional) URL to the proxy to be used for all API requests. URLs with "http", "https", and "socks5" schemes are supported. Can be sourced from `KUBE_PROXY_URL`.
 * `exec` - (Optional) Configuration block to use an [exec-based credential plugin](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins), e.g. call an external command to receive user credentials.
-* `api_version` - (Required) API version to use when decoding the ExecCredentials resource, e.g. `client.authentication.k8s.io/v1beta1`.
-* `command` - (Required) Command to execute.
-* `args` - (Optional) List of arguments to pass when executing the plugin.
-* `env` - (Optional) Map of environment variables to set when executing the plugin.
+  * `api_version` - (Required) API version to use when decoding the ExecCredentials resource, e.g. `client.authentication.k8s.io/v1beta1`.
```
BBBmau: These extra spaces should be removed.

BBBmau (changes requested):
When testing on my machine I was unable to get the behavior despite being able to run the tests and getting them to pass. @jrhouston was able to run manual tests on this PR and it all looks to be running as expected.

@n-oden We really appreciate your contribution! Just apply the documentation changes and we'll get this approved and merged 🎉

docs/index.md
n-oden: I agree! But `tfplugindocs generate` _really_ wants them to be there, and fixing this breaks the documentation check build.  I'll revert them out manually but someone should probably figure out why tfplugndocs keeps applying them.

BBBmau: yeah for right now we've been ignoring the documentation workflow due to it generating documentation that's just flat out wrong. Thanks for at least attempting to use it though!

BBBmau (approved):
This is a big contribution that the community has been asking for for quite sometime. You're more than welcome to pick up other issues that catch your eye if you're interested. Thank you again! 🎉

n-oden:
@BBBmau thank you for all of your help along this long and winding road!  Now... ETA to the next release of the provider? 🤣

BBBmau:
@n-oden Thank you for your contribution 😄 

https://github.com/hashicorp/terraform-provider-helm/releases/tag/v2.15.0

PR #1469, opened 2024-08-28
Creating bool debug

JaylonmcShan03:
### Description

This PR refactors the handling of the debug mode in the Helm provider to use a boolean variable instead of repeated string comparisons.

### Release Note

```release-note
...
```
### Community Note

### DIFF
--- a/helm-framework/helm/provider.go
+++ b/helm-framework/helm/provider.go
@@ -76,22 +76,22 @@ type KubernetesConfigModel struct {
 	Username              types.String `tfsdk:"username"`
 	Password              types.String `tfsdk:"password"`
 	Insecure              types.Bool   `tfsdk:"insecure"`
-	TlsServerName         types.String `tfsdk:"tls_server_name"`
+	TLSServerName         types.String `tfsdk:"tls_server_name"`
 	ClientCertificate     types.String `tfsdk:"client_certificate"`
 	ClientKey             types.String `tfsdk:"client_key"`
-	ClusterCaCertificate  types.String `tfsdk:"cluster_ca_certificate"`
+	ClusterCACertificate  types.String `tfsdk:"cluster_ca_certificate"`
 	ConfigPaths           types.List   `tfsdk:"config_paths"`
 	ConfigPath            types.String `tfsdk:"config_path"`
 	ConfigContext         types.String `tfsdk:"config_context"`
 	ConfigContextAuthInfo types.String `tfsdk:"config_context_auth_info"`
 	ConfigContextCluster  types.String `tfsdk:"config_context_cluster"`
 	Token                 types.String `tfsdk:"token"`
-	ProxyUrl              types.String `tfsdk:"proxy_url"`
+	ProxyURL              types.String `tfsdk:"proxy_url"`
 	// Exec                  types.List   `tfsdk:"exec"`
 }
 
 type ExecConfigModel struct {
-	ApiVersion types.String `tfsdk:"api_version"`
+	APIVersion types.String `tfsdk:"api_version"`
 	Command    types.String `tfsdk:"command"`
 	Env        types.Map    `tfsdk:"env"`
 	Args       types.List   `tfsdk:"args"`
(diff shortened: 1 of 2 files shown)


helm-framework/helm/provider.go
```diff
@@ -579,7 +583,7 @@ func (p *HelmProvider) Configure(ctx context.Context, req provider.ConfigureRequ
 
 	meta := &Meta{
 		Data: &HelmProviderModel{
-			Debug:                types.BoolValue(debug == "true"),
```
BBBmau, line 582: I believe there was one other place that had it set as `debug == "true"`. Could you double check just in case? I may just be misremembering also.

helm-framework/helm/provider.go
```diff
@@ -339,9 +338,14 @@ func (p *HelmProvider) Configure(ctx context.Context, req provider.ConfigureRequ
 	if resp.Diagnostics.HasError() {
 		return
 	}
+
+	debug := false
+	if os.Getenv("HELM_DEBUG") == "true" {
+		debug = true
+	}
 	// Override environment variables if the configuration values are provided
```
BBBmau: I'm curious how the config overrides the environment variables if both cases set the variable to true.

1. Getting the `HELM_DEBUG` where it's "true" will mark `debug = true
2. Getting the configuration values will mark `debug = true`

Yet i don't see how we apply the override of config values instead of env variables if:
> configuration values are set AND `HELM_DEBUG` envvar is set to "true"

helm-framework/helm/provider.go
JaylonmcShan03: Just did a GREP and didn't see it used anywhere else in the codebase. Also I believe it would throw an error if it was somewhere else I believe!

helm-framework/helm/provider.go
JaylonmcShan03: the logic sets debug to true if either the environment variable HELM_DEBUG is "true" or if the configuration explicitly sets debug to true. 

with my understanding this ensures that debugging is enabled if either source requests it.

BBBmau: In that case we should consider treating the two conditionals as one:
https://github.com/hashicorp/terraform-provider-helm/blob/393479e17b5d30b028690543568dc68c2143060b/helm-framework/helm/provider.go#L342-L348

can be switched to:
```hcl
debug := false
 if os.Getenv("HELM_DEBUG") == "true" || !config.Debug.IsNull() {
 	debug = true 
}
```

JaylonmcShan03: Gotcha! Due to the previous confusion where I treated two conditions as one, it wasn't very clear what was going on. But these checks are self explanatory so I will update it!

BBBmau: At a second glance I'm wondering if it's right to use `!config.Debug.IsNull()`

Since from my understanding if a user sets `debug = false` it is technically not null anymore and would return true and mark `debug = true` despite debug being set to false in the config. Thoughts?

Edit: This might be better suited: https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework@v1.11.0/types/basetypes#BoolValue.ValueBool

JaylonmcShan03: Good catch, you are right if the user sets debug. = false, due to it check if its not null it was simply just override it to true. I believe in order to properly handle this scenario > we should check both the null state and the actual boolean value of the Debug field. 

I believe it should look something like this > 
`debug := false

	if os.Getenv("HELM_DEBUG") == "true" {
		debug = true
	}

	if !config.Debug.IsNull() {
		debug = config.Debug.ValueBool()
	}`

BBBmau: i believe we can still keep them as one condition since from the docs it says:
>ValueBool returns the known bool value. If Bool is null or unknown, returns false. 

So in other words it would return the "known bool value" so it would return `true` if `debug = true` and `false` if `debug = false` in config.

if the value is `unknown` or `Null` would just return `false`.

helm-framework/helm/provider.go
```diff
@@ -339,10 +338,9 @@ func (p *HelmProvider) Configure(ctx context.Context, req provider.ConfigureRequ
 	if resp.Diagnostics.HasError() {
 		return
 	}
-	// Override environment variables if the configuration values are provided
-	if !config.Debug.IsNull() {
-		debug = fmt.Sprintf("%t", config.Debug.ValueBool())
-	}
+
+	debug := os.Getenv("HELM_DEBUG") == "true" || config.Debug.ValueBool()
```
BBBmau: nice one liner! One last request

Here `debug` is initialized on line 342 but is first used here in the `provider.go` file:
https://github.com/hashicorp/terraform-provider-helm/blob/2eb62ef550dbb1858ef03248b93fc6f53e98e36d/helm-framework/helm/provider.go#L460

It's best practice to include the variable initialization near the first call of the variable. In order to reduce the need to go searching for where the variable was first set.

BBBmau (approved):
nice red diff! 🎉

PR #1490, opened 2024-09-18
`framework-migration`: m.lock removal and more

BBBmau:
### Description

removes `m.Lock` from use since it is unnecessary after applying the necessary test changes that test concurrency

also resolves logic found that recomputesMetadata regardless of state (state being nil)

### Acceptance tests

### Release Note

```release-note
...
```
### Community Note


BBBmau:
From meeting: the recomputeMetadata issue comes from the logic not matching the expected behavior. This is the PR that does that refactor: https://github.com/hashicorp/terraform-provider-helm/pull/1471/files#diff-28289731fbd7b32bbd25297fea49c26bc344361e2d5ac15917d012f329d8bf1eR1984-R2005

The comment left says that we should `recompute the metadata when we find changes between the two values (plan and state)` yet the code returns true when both values are found to be equal which doesn't match the original logic at all. Latest commit addresses this issue and also adds a nil checker which returns true when `state == nil`

JaylonmcShan03 (approved):
After testing yesterday, I agree the lock should be removed. 
I tried the changes made to recomputeMetadata and it is now working!!!

Issue #989, opened 2022-11-10
OCI: The specified item already exists in the keychain, Still


jeremymcgee73:
### Terraform, Provider, Kubernetes and Helm Versions

```
Terraform version: 1.1.7
Provider version: 3.7.1
Kubernetes version: 1.21
```

### Affected Resource(s)

- helm_release

### Terraform Configuration Files
```hcl
data "aws_ecr_authorization_token" "token" {}

locals {
  helm_base_registry_path = trimprefix(data.aws_ecr_authorization_token.token.proxy_endpoint, "https://")
}

resource "helm_release" "example1" {
  name    = "example1"
  chart   = "oci://${local.helm_base_registry_path}/example1"
  version = "1.0.0"

  repository_username = data.aws_ecr_authorization_token.token.user_name
  repository_password = data.aws_ecr_authorization_token.token.password
}

resource "helm_release" "example2" {
  name    = "example2"
  chart   = "oci://${local.helm_base_registry_path}/example2"
  version = "2.0.0"

  repository_username = data.aws_ecr_authorization_token.token.user_name
  repository_password = data.aws_ecr_authorization_token.token.password
}
```

### Debug Output

```
Error: could not login to OCI registry "XXXXXXXX.dkr.ecr.us-east-1.amazonaws.com": error storing credentials - err: exit status 1, out: `The specified item already exists in the keychain.`
```

### Steps to Reproduce

1. `terraform apply`

### References

- #845

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

jeremymcgee73:
I'm still running into this, I believe a fix would be to check if the credential already exists. We could also create a new credential based on the resource name instead. This happens when we run multiple helm releases in parallel, with OCI.

jeremymcgee73:
This may also be OSX related only. I should be able to let you know if thats the case soon..

jeremymcgee73:
Also, this may only be a problem on OSX.

jeremymcgee73:
Just thought this would be helpful, tested with `2.8.0` and still had the same problem on OSX.

jeremymcgee73:
This only happens on OSX BTW. I have not seen it on Centos/AL2.

MikiLoz92:
[...]

PR #1535, opened 2024-11-15
GH-989: Handle osxkeychain rejecting the login request spuriously

ascopes:
This change implements a workaround for GH-989 where users on MacOS will occasionally see spurious issues with the `osxkeychain` rejecting an OCI registry login request.

The issue itself is likely a problem with upstream libraries, but until a time where this issue is fixed, this should work around the issue described in GH-989 to unblock builds.

### DIFF
--- a/helm/provider.go
+++ b/helm/provider.go
@@ -9,6 +9,7 @@ import (
 	"log"
 	"net/url"
 	"os"
+	"reflect"
 	"strconv"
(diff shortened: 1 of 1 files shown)


ascopes:
bump

ascopes:
inactive PRs awaiting review get closed now..?

PR #1537, opened 2024-11-18
Update CODEOWNERS

iBrandyJackson:
### Description

update CODEOWNERS to HybridCloud from legacy K8s team


SarahFrench (commented):
Looks like this team needs to be given permissions on the repo for this CODEOWNERS to be valid; that team isn't present in https://github.com/hashicorp/terraform-provider-helm/settings/access

iBrandyJackson:
@SarahFrench good catch, I've added the team ✅

SarahFrench (approved):

PR #1644, opened 2025-06-19
Fix sensitive values being replaced with '(sensitive value)' in Helm deployments

abatilo:
When using set_sensitive in the helm provider, the actual values deployed to
Kubernetes were incorrectly replaced with the string '(sensitive value)' instead
of the real sensitive value.

Root cause:
The logValues() function was using maps.Clone() which only creates a shallow
copy. Since Helm values are nested maps (e.g., {'configmap': {'foo': 'test'}}),
the inner maps were shared between the original and clone. When cloakSetValues()
modified the 'cloned' map to mask sensitive values for logging, it was actually
modifying the original map that was then passed to Helm for deployment.

Fix:
- Implemented deepCloneMap() function that recursively clones nested maps
- Updated logValues() to use deepCloneMap instead of maps.Clone
- Updated setReleaseAttributes() to use deepCloneMap for consistency
- Removed unused 'maps' import

This ensures sensitive values are properly masked in logs and Terraform state
display while the actual values are correctly deployed to Kubernetes.

Fixes the issue where ConfigMaps would contain '(sensitive value)' instead of
the actual sensitive data when using set_sensitive blocks.

### DIFF
--- a/.changelog/1644.txt
+++ b/.changelog/1644.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+`helm_release`: Fix shallow clone bug causing nested sensitive values to be redacted in the k8s API
+```
--- a/helm/data_helm_template_test.go
+++ b/helm/data_helm_template_test.go
@@ -22,11 +22,12 @@ func TestAccDataTemplate_basic(t *testing.T) {
 		Steps: []resource.TestStep{{
 			Config: testAccDataHelmTemplateConfigBasic(testResourceName, namespace, name, "1.2.3"),
 			Check: resource.ComposeAggregateTestCheckFunc(
-				resource.TestCheckResourceAttr(datasourceAddress, "manifests.%", "5"),
+				resource.TestCheckResourceAttr(datasourceAddress, "manifests.%", "6"),
 				resource.TestCheckResourceAttrSet(datasourceAddress, "manifests.templates/deployment.yaml"),
 				resource.TestCheckResourceAttrSet(datasourceAddress, "manifests.templates/service.yaml"),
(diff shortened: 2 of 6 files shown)


abatilo:
@jrhouston would you mind taking a look?

abatilo:
This PR fixes the `(sensitive value)` bug described here:
https://github.com/hashicorp/terraform-provider-helm/issues/1637#issuecomment-2986527876

jrhouston:
Thanks for jumping in to fix this @abatilo. I've pushed a reproducer test for this to your branch, which passes with your change. 

It looks like the code that got missed in the migration from SDKv2 is this janky JSON Unmarshal/Marshal step which I guess is a cheat way to do a deep copy. 

https://github.com/hashicorp/terraform-provider-helm/blob/afc64eda2181bf835ac648177e7464416c9b9fef/helm/resource_release.go#L1425-L1431

abatilo:
Of course @jrhouston , happy to try and contribute.

Sorry, can you help me understand what I should do next? I see some acceptance tests are failing. Should I be the one trying to fix those? Or are you taking over the branch, etc?

jrhouston:
My apologies, that test was failing because I added a secret to the test chart for testing the sensitive values but forgot to update one of the tests which asserts the number of templates in the chart. I've fixed the test and pushed it, everything should turn green now. 

I think this PR will be good to go, we will need a second approval from @jaylonmcshan19-x before I can hit merge.

jrhouston (approved):

abatilo:
Amazing. Thank you so much.

jgournet:
I know this does not help in any way, but:
Thank you very much for addressing this so fast !

jaylonmcshan19-x (approved):
Thanks for addressing this bug @abatilo @jrhouston :)

PR #1667, opened 2025-07-01
Handle Unknown `version` during `plan`

YannickAlex07:
### Description

During some of our projects we actually compute the correct version to apply during `apply`-time (our reasons are a bit odd but the final version is produced by another resource after pushing the packaged Helm chart).

This is probably a bit of a niche use case, however we wondered if it would be possible to consider keeping the `version` unknown if the input string is already unknown instead of raising an error due to version mismatches. If no version is specified, the current behavior should still apply.

### Acceptance tests

### Release Note

```release-note
NONE
```

### Community Note

### DIFF
--- a/helm/resource_helm_release.go
+++ b/helm/resource_helm_release.go
@@ -2036,7 +2036,9 @@ func (r *HelmRelease) ModifyPlan(ctx context.Context, req resource.ModifyPlanReq
 		plan.Version = types.StringNull()
 	}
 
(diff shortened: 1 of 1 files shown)


YannickAlex07:
Before I start to add any tests, I would appreciate an opinion if something like this would be acceptable or if the current behavior is wanted and the change would not be acceptable / make sense.

PR #1734, opened 2025-12-15
fix(helm_release): Preserve Terraform state on failed Helm operations (#1669)

### LINKED ISSUE #1669: v3 Provider appears to randomly remove helm_release resources from TF state - Hard to reproduce

```
Terraform version: 1.5.7
Provider version: 3.0.2
Kubernetes version: 1.31 - EKS
```

### Affected Resource(s)
- helm_release

N/A

### Expected Behavior
helm_release resources that have not been changed do not get their TF state files changed at all

### Actual Behavior
After several successful Terraform apply runs (no changes to any `helm_release` resources), the terraform state for several (but not all) `helm_release` resources was updated and those resources completely removed from the state.

This resulted in Terraform trying to recreate those Helm Releases even though they already existed in the EKS cluster, resulting in an error : 

`cannot re-use a name that is still in use`

### Important Factoids
We cannot debug the whole thing and we are having trouble easily reproducing this as we are seeing the outcomes and comparing terraform states (v1 is ok and then v2 is empty).
We are also a bit suspicious of this method: https://github.com/hashicorp/terraform-provider-helm/blob/main/helm/resource_helm_release.go#L873 especially the resourceReleaseExists return that removes resources from the context below.

desek:
## Rollback Plan

If a change needs to be reverted, we will publish an updated version of the library.

## Changes to Security Controls

No changes to security controls (access controls, encryption, logging) in this pull request.

### Description

This PR fixes a critical bug where `helm_release` resources are randomly removed from Terraform state after failed deployments or during refresh operations. This issue causes subsequent `terraform apply` runs to fail with the error "cannot re-use a name that is still in use" because Terraform attempts to recreate releases that already exist in the Kubernetes cluster but are no longer tracked in state.

**Root Causes Addressed:**

1. **Update Function (Primary Fix)**: When a Helm upgrade fails, the function now saves state before returning an error. Previously, it returned immediately without updating state, causing state loss when combined with subsequent Read operations.

2. **Create Function**: Added state persistence before returning error on failed create to prevent orphaning releases from Terraform tracking.

3. **Read Function**: Improved error handling order and added informative logging when removing resources from state.

4. **resourceReleaseExists Function**: Completely rewritten to use `action.List` instead of `getRelease` to detect releases in ALL states (deployed, failed, pending-install, pending-upgrade, etc.). This is more comprehensive and prevents false negatives that led to state removal.

**Changes Summary:**
- `helm/resource_helm_release.go`: Enhanced error handling in Create, Read, and Update functions to preserve state on failures; rewrote `resourceReleaseExists` for comprehensive release detection
- `helm/resource_helm_release_test.go`: Added 8 new acceptance tests covering various failure scenarios

Fixes #1669

### Acceptance tests

**New Acceptance Tests:**
- `TestAccResourceRelease_updateExistingFailed` - Tests failed deployment preserves state and recovery
- `TestAccResourceRelease_statePreservedDuringRefresh` - Tests state not removed during refresh
- `TestAccResourceRelease_refreshPreservesFailedState` - Tests refresh preserves failed release state
- `TestAccResourceRelease_comprehensiveReleaseDetection` - Tests release detection in all states
- `TestAccResourceRelease_failedInitialDeployPreservesState` - Tests failed initial deployment preserves state
- `TestAccResourceRelease_failedInitialDeployAtomicNoState` - Tests atomic=true behavior on failed installation
- `TestAccResourceRelease_deleteOperationCorrectBehavior` - Tests delete operation maintains correct behavior
- `TestAccResourceRelease_deleteAlreadyRemovedRelease` - Tests delete handles already-removed release gracefully

### Release Note

```release-note
bug: Fix helm_release resources being randomly removed from Terraform state after failed deployments or during refresh operations (#1669)
```
### References

- GitHub Issue: #1669

### Community Note

### DIFF
--- a/helm/resource_helm_release.go
+++ b/helm/resource_helm_release.go
@@ -968,6 +968,9 @@ func (r *HelmRelease) Create(ctx context.Context, req resource.CreateRequest, re
 			return
 		}
 
+		// Save state to prevent orphaning the release from Terraform tracking
+		resp.Diagnostics.Append(resp.State.Set(ctx, &state)...)
+
 		resp.Diagnostics.Append(diag.NewWarningDiagnostic("Helm release created with warnings", fmt.Sprintf("Helm release %q was created but has a failed status. Use the `helm` command to investigate the error, correct it, then run Terraform again.", client.ReleaseName)))
 		resp.Diagnostics.Append(diag.NewErrorDiagnostic("Helm release error", err.Error()))
 
@@ -1015,17 +1018,18 @@ func (r *HelmRelease) Read(ctx context.Context, req resource.ReadRequest, resp *
 		return
 	}
 
+	logID := fmt.Sprintf("[resourceReleaseRead: %s]", state.Name.ValueString())
+
 	exists, diags := resourceReleaseExists(ctx, state.Name.ValueString(), state.Namespace.ValueString(), meta)
-	if !exists {
-		resp.State.RemoveResource(ctx)
-		return
-	}
 	resp.Diagnostics.Append(diags...)
 	if resp.Diagnostics.HasError() {
 		return
 	}
-
-	logID := fmt.Sprintf("[resourceReleaseRead: %s]", state.Name.ValueString())
+	if !exists {
+		tflog.Info(ctx, fmt.Sprintf("%s Release not found in namespace %s, removing from Terraform state. If this is unexpected, the release may have been deleted outside of Terraform or may be in a state that cannot be detected.", logID, state.Namespace.ValueString()))
+		resp.State.RemoveResource(ctx)
+		return
+	}
 	tflog.Debug(ctx, fmt.Sprintf("%s Started", logID))
 
 	c, err := meta.GetHelmConfiguration(ctx, state.Namespace.ValueString())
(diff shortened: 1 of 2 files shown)


desek:
@jrhouston / @jaylonmcshan19-x - any chance you can have a look at this?

CRASH-Tech:
Merge please

schnell3526:
Hi @desek, thank you for this important fix! I've cherry-picked your changes into my fork and found a few issues in the test cases that cause CI failures:

### 1. `TestAccResourceRelease_updateExistingFailed`

The expected revision in Step 4 should be `"4"` instead of `"3"`, because each failed upgrade (Step 2 and Step 3) increments the Helm revision:
- Step 1: revision 1 (initial deploy)
- Step 2: revision 2 (failed upgrade)
- Step 3: revision 3 (failed upgrade again)
- Step 4: revision 4 (successful recovery)

### 2. `TestAccResourceRelease_statePreservedDuringRefresh` and `TestAccResourceRelease_refreshPreservesFailedState`

These tests fail with:
```
TestStep cannot have Config or ConfigDirectory or ConfigFile and RefreshState
```

The `terraform-plugin-testing` framework does not allow `Config` and `RefreshState` to be used in the same `TestStep`. The fix is to remove `Config` from the `RefreshState` steps.

### 3. Hardcoded `"FAILED"` status

Several tests use hardcoded `"FAILED"` string, but Helm returns lowercase `"failed"`. The fix is to use `release.StatusFailed.String()` for consistency with other tests in this file.

---

I've fixed these issues in my fork. Here are the commits if they help:
- https://github.com/schnell3526/terraform-provider-helm/commit/802470130
- https://github.com/schnell3526/terraform-provider-helm/commit/da816b8bd

desek:
@schnell3526 can you have a look if it passes the test now?

desek:
@schnell3526 any timeline on merging this?

desek:
With the recently changes CODEOWNERs. @iam404 / @rigalGit - can you have a look at this?

bitte-ein-bit:
I arrived at the same `Read` change independently while debugging a state loss on AKS, found this PR afterwards, and dropped that hunk from mine. Some evidence in support of it, since I think this fixes more than the description claims.

The reorder is not only about failed deployments. `resourceReleaseExists` returns `false` both when the release is absent and when the lookup failed, and the two are only distinguishable through the diagnostics. With the original ordering the diagnostics are discarded, so **any transient error talking to the API server is indistinguishable from a deleted release** and the release is removed from state.

That is what happened to us. In a single terragrunt run against one AKS cluster, the `kubernetes` provider reported

```
Error: Invalid configuration for API client
Get "https://<redacted>:443/apis": net/http: TLS handshake timeout
```

and was retried successfully, while the `helm` provider silently dropped a `helm_release` from state during refresh. The next plan showed `1 to add` and the apply failed with:

```
Error: installation failed
cannot re-use a name that is still in use
```

That error only comes from `Install.availableName()` when `Releases.History(name)` returns a live revision, so the release existed in cluster storage the entire time — the refresh result was wrong, not the cluster state. Helm's release storage reads the release Secrets through the same API server that was timing out, and `Get.Run` starts with `KubeClient.IsReachable()`, so the failure lands squarely in `getRelease`.

This is probably also why #1669 is so hard to reproduce: it needs the API server to blip during a refresh, independent of chart or configuration. Might be worth mentioning in the PR description — it makes the case for this change quite a bit stronger than "state loss after failed deployments".

Two things I noticed while comparing:

**1. `Delete` has the same ordering and is not covered here.** A failed lookup there is read as "already uninstalled", so it returns early, the resource leaves state, and the uninstall never runs — the release keeps running in the cluster untracked. I've opened #1867 for just that; it deliberately does not touch `Read` so it does not conflict with this PR, and the two can merge in either order.

**2. I don't think the `resourceReleaseExists` rewrite is needed, and it may narrow detection slightly.** Line references are against `helm.sh/helm/v3@v3.20.2`, the version currently pinned in `go.mod`.

The premise is that `action.Get` "may not return releases in certain states", but it applies no status filter at any point:

* `Get.Run` → `cfg.releaseContent(name, 0)` → `cfg.Releases.Last(name)` (`pkg/action/action.go:314-315`)
* `Storage.Last` → `History(name)`, reverse-sort by revision, return the newest — no status logic (`pkg/storage/storage.go:231-243`)
* `Storage.History` → `Driver.Query(map[string]string{"name": name, "owner": "helm"})` — no `status` label (`pkg/storage/storage.go:153-157`)

Compare `Storage.DeployedAll`, which *does* add `"status": "deployed"` (`pkg/storage/storage.go:134-149`) — that is the query behind "has no deployed releases", and it isn't on this path. So failed, pending-install, pending-upgrade and uninstalled-with-history releases already resolve through `getRelease` today.

The status filtering people run into on the CLI (`helm list` needing `-a`/`--pending`) comes from `action.NewList` defaulting to `StateMask: ListDeployed | ListFailed` (`pkg/action/list.go:142`). That is a property of `List`, not of `Get`.

Going through `List` also has one edge that `Get` does not: `ListAll` omits `ListUnknown` (`pkg/action/list.go:84`, constant at `:53`), and `filterStateMask` drops anything the mask doesn't match, mapping the status string through `FromName`, which returns `ListUnknown` for anything unrecognized (`pkg/action/list.go:264-277` and `:80`). Since `Status` is a plain string whose zero value is `""` (`pkg/release/status.go:44`), a release whose newest revision has status `unknown` or an empty status field would come back as non-existent — re-triggering the state removal this PR is fixing. Rare in practice, since Helm doesn't normally persist `unknown`, but it points the wrong way.

Dropping that hunk and keeping `getRelease` would make the diff smaller and easier to review, and the ordering fix alone resolves #1669. Entirely your call though — it doesn't block anything, and I'd rather see this merged as-is than stall on it.

Zhann:
Deterministic reproduction of the Read path this PR fixes, verified red on 3.2.0 and green with this PR's reordering: https://github.com/hashicorp/terraform-provider-helm/issues/1669#issuecomment-5816305266.

PR #1749, opened 2026-02-02
Fix StateUpgrader 'unsupported attribute resources' error

### LINKED ISSUE #18: Outputs..
Hiya,

Just wondering how we can get outputs from charts for use with other terraform modules?

Say we deploy the influxdb helm chart - what would be the right syntax to use to get an output variable - like the IP address?

schnell3526:
## Summary

- Add `resources` attribute to `oldType` in version 0 StateUpgrader
- Add `resources` attribute to `oldType` in version 1 StateUpgrader
- Add unit tests to verify the fix

## Problem

When upgrading state that contains a `resources` attribute (from schnell3526/helm provider), the StateUpgrader failed with:

```
Failed to marshal state to json: unsupported attribute "resources"
```

This occurred because the `oldType` definition in the StateUpgrader did not include the `resources` attribute. When tftypes attempts to unmarshal a state that contains an attribute not defined in the type, it fails with "unsupported attribute" error.

## Solution

Add `resources: tftypes.Map{ElementType: tftypes.String}` to `oldType` in both version 0 and version 1 StateUpgraders.

This allows the StateUpgrader to correctly handle:
- State WITH resources attribute (from schnell3526/helm provider) - now works
- State WITHOUT resources attribute (from hashicorp/helm provider) - continues to work

## Test plan

- [x] Added unit tests for StateUpgrader oldType unmarshal behavior
- [x] `go test -v -run TestStateUpgrader ./helm/...` passes

Closes #18

🤖 Generated with [Claude Code](https://claude.ai/code)


schnell3526:
Closing - PR was created to the wrong repository by mistake.

PR #1816, opened 2026-07-27
Fix #1788: Handle escaped dots (\.) in set_sensitive key paths

### LINKED ISSUE #1788: set_sensitive fails to redact values when key name contains escaped dots (\.)
### OpenTofu, Provider, Kubernetes and Helm Versions

```
OpenTofu version: 1.10.6
Provider version: 3.1.1
```

### Affected Resource(s)
- helm_release

### Description

This is a re-report of #737 (closed) and is also related to PR #746 (never merged). The bug still exists in provider v3.1.1.

When using `set_sensitive` with a key name that contains a literal `.` (escaped with `\\.` in HCL), the sensitive value is **not redacted** in the `metadata` attribute. The value appears in plaintext in plan output.

### Root Cause

`cloakSetValue()` in `resource_helm_release.go` uses `strings.Split(valuePath, ".")` which splits on **all** dots, including backslash-escaped ones. Given a path like `server.config.oidc\.config`, it produces `["server", "config", "oidc\", "config"]` (4 segments) instead of `["server", "config", "oidc.config"]` (3 segments). Since Helm's `strvals.ParseInto()` correctly handles the escaping when setting the value into the map, the cloaking function can't find the key to mask.

The same bug exists in `cloakSetValueModel()` in `data_helm_template.go`.

```hcl
  name       = "my-release"
  repository = "https://charts.example.com"
  chart      = "my-chart"

  set_sensitive {
    name  = "config.oidc\\.clientSecret"
    value = var.oidc_client_secret
  }
}
```

### Steps to Reproduce
1. `tofu apply` with a `set_sensitive` block where the name contains `\\.`
2. `tofu plan` or `tofu destroy` — observe the sensitive value in plaintext under `metadata.values`

### Expected Behavior

The value should appear as `(sensitive value)` in the metadata output:
```
metadata.values = jsonencode({
  config = {
    "oidc.clientSecret" = "(sensitive value)"
  }
})
```

### Actual Behavior

The actual secret value is displayed in plaintext in the metadata output.

### References
- #737 (original issue, closed)
- #746 (fix PR, never merged)
- #1221 (related — sensitive values leaking from `values` attribute)

dennismdejong:
## Description

Fixes #1788. When `set_sensitive` has a key path containing backslash-escaped dots (`\.`), the value is not properly redacted in logs/state.

**Root cause:** `cloakSetValue()` and `cloakSetValueModel()` use `strings.Split(valuePath, ".")` which splits on ALL dots including escaped ones. For a path like `config.oidc\.clientSecret`, it was incorrectly split into `["config", "oidc\", "clientSecret"]` instead of `["config", "oidc.clientSecret"]`.

**Fix:** Added `splitPath()` function that respects backslash-escaped dots, treating `\.` as a literal dot in the key name rather than a path separator.

## Changes
- Added `splitPath()` function in `helm/resource_helm_release.go` that splits on unescaped dots only
- Updated `cloakSetValue()` and `cloakSetValueModel()` to use `splitPath()` instead of `strings.Split()`
- Uncommented and updated existing unit tests for `cloakSetValue()`, added new test for escaped dots

### DIFF
--- a/helm/data_helm_template.go
+++ b/helm/data_helm_template.go
@@ -1118,7 +1118,7 @@ func cloakSetValuesModel(config map[string]interface{}, state *HelmTemplateModel
 const sensitiveContentModelValue = "(sensitive value)"
 
 func cloakSetValueModel(values map[string]interface{}, valuePath string) {
-	pathKeys := strings.Split(valuePath, ".")
+	pathKeys := splitPath(valuePath)
 	sensitiveKey := pathKeys[len(pathKeys)-1]
 	parentPathKeys := pathKeys[:len(pathKeys)-1]
 
--- a/helm/data_helm_template.go
+++ b/helm/data_helm_template.go
@@ -886,6 +886,9 @@ func chartPathOptionsModel(model *HelmTemplateModel, meta *Meta, cpo *action.Cha
 	}
 
 	version := getVersionModel(model)
+	if registry.IsOCI(repository) {
+		version = strings.TrimPrefix(version, "v")
+	}
 
 	cpo.CaFile = model.RepositoryCaFile.ValueString()
 	cpo.CertFile = model.RepositoryCertFile.ValueString()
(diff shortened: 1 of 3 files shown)


shadan227:
Commenting here for traction, hope this gets resolved soon.

PR #1873, opened 2026-08-20
Migrate to Helm v4 SDK

### LINKED ISSUE #1730: Support for Helm v4
### Description

On Nov 12 2025 Helm team released a new major version, v4.
At this stage, there aren't many breaking end-user features, but the API has been bumped to v4 (see [helm.sh/helm/v4](https://pkg.go.dev/helm.sh/helm/v4)). At some point, users will likely start migrating to v4 and will expect the provider to support it.

### Potential Terraform Configuration

```hcl
resource "helm_release" "loki" {
  repository = "https://grafana.github.io/helm-charts"
  chart      = "loki"
  version    = "sha256:abc123..." # supports pinning by digest now
  name        = "loki"
  namespace   = kubernetes_namespace_v1.this.metadata[0].name
  # Deprecated
  # atomic      = true 
  rollback_on_failure = true
  # Deprecated
  # force_update = true
  force_replace = true

  values = [file("${path.module}/values.yaml")] # supports multi-document YAML now
  # Removed, postrenderers are plugins now
  # postrender {...}
  ...
}

```

### References

[https://helm.sh/docs/overview](https://helm.sh/docs/overview)
[https://pkg.go.dev/helm.sh/helm/v4#section-readme](https://pkg.go.dev/helm.sh/helm/v4#section-readme)

### Community Note

### LINKED ISSUE #1792: Helm 4 support
### Description

Add Support for Helm 4.

It was release some time ago and I'm interested in server side apply which is default behaviour in Helm 4.

### References

https://helm.sh/docs/overview/#server-side-apply

### Community Note

dennismdejong:
## Rollback Plan

If a change needs to be reverted, we will publish an updated version of the library.

## Changes to Security Controls

No changes to security controls.

### Description

This PR migrates the provider from the Helm v3 SDK (`helm.sh/helm/v3`) to the Helm v4 SDK (`helm.sh/helm/v4` v4.2.3), resolving issues #1730 and #1792.

### What changed

- **Dependency bump**: `helm.sh/helm/v3` → `helm.sh/helm/v4` v4.2.3; `k8s.io` deps upgraded to v0.36.1
- **Import paths**: All 31 `helm.sh/helm/v3/...` imports updated to `helm.sh/helm/v4/...` across 6 source files + tests
- **Removed packages**: `chartutil`, `releaseutil`, `pkg/repo`, `pkg/postrender` no longer exist in v4:
  - Added `helm/compat_v4.go` shims for `ParseKubeVersion`, `SplitManifests`, and dependency conversion
  - Added `helm/postrender_exec.go` to preserve binary-path post-renderer support (v3's `postrender.NewExec` was removed)
- **Action field changes** (v4 API):
  - `Wait` bool → `WaitStrategy` (`kube.WaitStrategy`)
  - `WaitForJobs` → handled by `StatusWatcherStrategy`
  - `Atomic` → `RollbackOnFailure`
  - `Force`/`ForceUpdate` → `ForceReplace`
  - `DryRun` → `DryRunStrategy`
  - `ClientOnly` → `DryRunStrategy`
  - Removed `Recreate` (removed in Helm v4)
- **Type assertions**: `loader.Load` returns `chart.Charter` (→ `*v2.Chart`); action `Run()` returns `release.Releaser` (→ `*v1.Release`)
- **Config**: `actionConfig.Init` no longer accepts a logging callback
- **Tests**: Updated test helpers (`kube.GetConfig` → `ConfigFlags`, `StatusDeployed` → `release/common`, status action fields)

### Breaking changes for users

- Bundled Helm engine is now v4 (server-side apply default, digest pinning, multi-doc YAML values)
- `postrender` binary executables are deprecated in favor of Helm v4 plugins (still supported via compat layer)
- `recreate_pods` is no longer supported

### Acceptance tests

(Note: This is a large migration. Acceptance tests require a live Kubernetes cluster and were not run here; the code compiles and passes `go vet`.)

### Release Note

```release-note
Migrate the provider to the Helm v4 SDK
```

### References

Fixes #1730, Fixes #1792

### Community Note

* Please vote on this issue by adding a 👍 reaction to the original issue to help the community and maintainers prioritize this request


dennismdejong:
## Acceptance test results

All acceptance tests now **pass** against a live Kubernetes cluster (minikube v1.38.1 with the podman driver, Kubernetes v1.35.1).

Ran the full `TestAcc` suite: **all tests pass** (`ok ... 241s`).

### What was validated
- **helm_release**: basic CRUD, values/set/set_list/literalSet/set_wo, postrender, update, upgrade_install (cold+warm start), dependency, delete_regression, createNamespace, Timeouts, takeOwnership, identity, import, parallel, multiple_releases, chartURL, repository_url, helm_repo_add, LintFail, FailedDeploy, emptyVersion, inconsistentVersion
- **manifest experiment**: manifest, manifestUnknownValues, manifestServerDiff (fixed)
- **OCI**: OCI_repository, OCI_registry_login, OCI_login (fixed - now works with podman + plain-HTTP local registries)
- **helm_template data source**: basic, crds, templates, kubeVersion, configSetNull
- **state upgrader**: MetadataStructure, values, PostrenderStructure
- **unit**: TestConvertYAMLManifestToJSON

### Notable fixes made during test validation
- ** → ** (not ): in Helm v4, `ForceReplace` cannot be combined with server-side apply. Mapping `force_update` to `ForceConflicts` correctly enables drift recovery under SSA.
- **OCI plain-HTTP**: added `needsPlainHTTP` + a dedicated plain-HTTP registry client for local (localhost/127.0.0.1) OCI registries, fixing `LocateChart` pulls against local registries.
- **metadata.generation stripping**: removed `metadata.generation` from tracked resources to avoid spurious diffs under Helm v4 server-side apply.
- **valuesUnknown bug**: fixed a copy-paste bug where `set_list` values were read from `plan.Set`.
- **podman support**: `setupOCIRegistry` now prefers `podman` over `docker` for the local test registry.

dennismdejong:
Correction to the previous comment (the `force_update` → `ForceConflicts` line was garbled):

- **`force_update` → `ForceConflicts`** (not `ForceReplace`): in Helm v4, `ForceReplace` cannot be combined with server-side apply (error: "cannot use server-side apply and force replace together"). Mapping `force_update` to `ForceConflicts` correctly enables drift recovery under the default server-side apply behavior.

igor-nikiforov:
Is there any ETA of merging this?

helm/resource_helm_release.go
EronWright:
```suggestion
				if strings.Contains(strings.ToLower(err.Error()), "kubernetes cluster unreachable") {
```

Various error messages were lower-cased in v4 (there may be other instances): https://github.com/helm/helm/commit/3877ec9049b18694da05c29481f2b2191cd21802#diff-137bba79988de442c1fde0fc29520d6d7e5a5239ceac82f5439a5bf513ebb2d4L187-R187

EronWright (commented):
### 1. The cluster-unreachable escape hatch no longer matches

`resource_helm_release.go:2181`:

```go
if strings.Contains(err.Error(), "Kubernetes cluster unreachable") {
```

Helm lowercased that string in helm/helm@3877ec9 — collateral from a `golangci-lint` ST1005 fix:

https://github.com/helm/helm/commit/3877ec9049b18694da05c29481f2b2191cd21802#diff-137bba79988de442c1fde0fc29520d6d7e5a5239ceac82f5439a5bf513ebb2d4L187-R187


Worth fixing properly while you're here: that branch calls `AddError`, so it fails the plan either way, and sets `plan.Manifest = types.StringNull()` on a Computed attribute where `StringUnknown()` is meant. `AddWarning` + `StringUnknown` would make it do what the comment says.

Even better would be to implement deferral. The provider already defers in `Configure` (`provider.go:339`, `DeferredReasonProviderConfigUnknown`), and the framework carries `Deferred` on `ModifyPlanResponse` plus `DeferralAllowed` on the request. `terraform-provider-kubernetes` does exactly this in `manifest/provider/plan.go:305`: defers with `DeferredReasonResourceConfigUnknown` on a no-match GVK *and* downgrades the diagnostic to a warning. The same shape here would cover both this case and the CRD-not-yet-installed one, and it would replace error-string matching with a protocol signal.

### 2. `WaitForJobs` still exists in v4 and isn't being set

The description says it's "handled by `StatusWatcherStrategy`", but it's still a field and still honoured — `install.go:99,546` and `upgrade.go:78,494`. Collapsing both flags into `wait || wait_for_jobs` changes behaviour in both directions, because v3 nested them (`install.go:469`: `if i.Wait { if i.WaitForJobs {...} }`):

| config | v3 | this PR |
|---|---|---|
| `wait=true, wait_for_jobs=true` | waits for jobs | **doesn't** — `WaitForJobs` never set |
| `wait=false, wait_for_jobs=true` | no wait at all | **waits for everything**, can hit `timeout` |

Keeping `WaitForJobs = plan.WaitForJobs.ValueBool()` alongside the strategy fixes both.

### 3. `recreate_pods` deprecation is partially-done

The CHANGELOG says "no longer supported", but the attribute is still in the schema (`:441`), still `Optional + Computed`, still defaulted, still described as *"Perform pods restart during upgrade/rollback"*. Only the two lines that read it were deleted. So `recreate_pods = true` parses, plans and applies clean while doing nothing.

A `DeprecationMessage` (warn) or a config validator (error) would make the break visible.

### 4. `force_update` description is now inaccurate

`ForceConflicts` is the correct mapping under SSA — agreed with your follow-up, and `ClientUpdateOptionForceReplace` is documented as mutually exclusive with SSA. But v3's `Force` meant *delete/recreate to get past an immutable field*, and `ForceConflicts` means *win a field-manager ownership dispute*. Users reaching for `force_update` to escape an immutable-field rejection will now silently get nothing, while the schema still promises:

```
Description: "Force resource update through delete/recreate if needed."
```

At minimum the description and CHANGELOG need updating, or separate into `force_conflicts` / `force_replace`.

### 5. Server-side apply lacks options

Nothing here sets `ServerSideApply`, so it comes in as a helm default — and the two defaults differ: `NewInstall` is `true` (`install.go:166`), `NewUpgrade` is `"auto"` (`upgrade.go:149`), which resolves from the prior release's `ApplyMethod` (`upgrade.go:313`). Releases created by the v3 provider therefore keep client-side three-way merge indefinitely, while anything created after this bump is SSA — a mixed-mode fleet with no attribute to observe or control it.

Given the CHANGELOG leads with "server-side apply as the default behavior", worth elaborating on the new behavior.

### 6. `valuesUnknown` — `set_list` decode is the wrong model

```go
setList := []setResourceModel{}
plan.SetList.ElementsAs(context.Background(), &setList, false)   // :2638
```

Good catch on the original copy/paste, but `set_list` elements are `{name = string, value = list(string)}` (`set_listResourceModel`), not `setResourceModel` (`{name, type, value}` all strings). The decode is a type mismatch, and since the returned diagnostics are discarded, `setList` stays empty and the loop never runs — so element-level unknowns in `set_list` are still undetected, now with an error being swallowed rather than a harmless redundant decode.

Decoding into `set_listResourceModel`, checking `Value.IsUnknown()` plus its elements, and appending the diags would close it.

### 7. `plan.Metadata = state.Metadata` — what is this fixing?

```go
} else if state != nil && !plan.Metadata.IsUnknown() && !state.Metadata.IsNull() && !state.Metadata.IsUnknown() {
    plan.Metadata = state.Metadata   // :2378
}
```

As far as I can tell this is dead: `metadata` is Computed-only, so the framework already copies prior state into the plan, and every path in `ModifyPlan` that touches it sets `ObjectUnknown` — which the `!plan.Metadata.IsUnknown()` guard excludes. If it isn't dead, then it's forcing the plan to claim metadata is unchanged when something has decided otherwise, which would trade a spurious diff for an inconsistent-result-after-apply.

Could you say which failure motivated it? Same question for `delete(md, "generation")` (`:2711`) — defensible on its own, but it arrives unexplained in a migration PR, which makes it hard to tell whether it's fixing v4 fallout or papering over it.

### 8. `OCIRegistryLogin`: registry client dropped on the non-OCI path

Moving `actionConfig.RegistryClient = registryClient` below the `if ociURL == "" { return diags }` early return means it's now nil for classic HTTP repos and local chart paths — `ModifyPlan` calls this unconditionally. Was that intended? It reads as unrelated to the v4 port.

Same file, `needsPlainHTTP`: hardcoding `localhost`/`127.0.0.1`/`::1` → plain HTTP silently downgrades a TLS registry reached over loopback (port-forward is the common case) and doesn't help a plain-HTTP registry on any other host, with no way to opt out either direction. Feels like it wants to be a provider attribute, and like it belongs in its own PR.

### Nits

- `compat_v4.go` reimplements something that's already exported from a package the file imports: `util.BySplitManifestsOrder` lives at `pkg/release/v1/util/manifest.go:73`, right next to the `util.SplitManifests` you're calling. The hand-rolled `strconvAtoi` can then go too — it returns `(0, nil)` on a parse error, silently mis-ordering manifests, where `strconv.Atoi` would not.
- The five `loader.Load(...).(*v2.Chart)` assertions will fail with *"Could not type assert chart to v2.Chart"* the moment chart v3 is exported — `pkg/chart/common.go:37` already type-switches on it via `internal/chart/v3`. Routing through `chart.NewAccessor` where possible would age better.
- A few `fmt.Sprintf` calls with no formatting verbs (e.g. the `getChartModel` error).

helm/resource_helm_release.go
```diff
@@ -1131,29 +1168,38 @@ func (r *HelmRelease) Update(ctx context.Context, req resource.UpdateRequest, re
...
+		client.WaitStrategy = kube.HookOnlyStrategy
+	}
+	client.DryRunStrategy = action.DryRunNone
 	client.DisableHooks = plan.DisableWebhooks.ValueBool()
-	client.Atomic = plan.Atomic.ValueBool()
+	client.RollbackOnFailure = plan.Atomic.ValueBool()
 	client.SkipCRDs = plan.SkipCrds.ValueBool()
 	client.SubNotes = plan.RenderSubchartNotes.ValueBool()
 	client.DisableOpenAPIValidation = plan.DisableOpenapiValidation.ValueBool()
-	client.Force = plan.ForceUpdate.ValueBool()
```
EronWright, line 1153: Since the SSA mode is 'auto' for updates, the `Force` flag might still have significance in some cases.

Issue #1791, opened 2026-05-19
New resources attribute in 3.1.0 causes drift on deployment image specs


patrickvinograd:
Somewhere between a bug and a question:
* We have a large terraform project with multiple charts/instances of `helm_release` representing our various app deployments.
* The charts for these apps will have a template that refers to `<app_name>:latest` in the deplyment's image spec.
* We also continuously deploy these apps from our CI pipeline by patching the image to `<app_name>:<hash>. 

While it's true that we have both helm/terraform and this external CI process managing the same resource, up until helm provider 3.0.2, there was no conflict here. 

But upon upgrading to 3.1.1, we see essentially continuous drift in these helm_release resources, where any `plan` operation wants to change:
```
resources                  = {
     ....
        spec     = {
            ~ containers                    = [
                ~ {
                    ~ image                    = "myapp:fde8e913f8" -> "myapp:latest"
...
```
I.e. to restore the image spec back to `latest` as specified in the chart.

We've narrowed down this behavior change to https://github.com/hashicorp/terraform-provider-helm/pull/1693, which adds the `resources` attribute to helm_release, and that is indeed the attribute that is showing as drifted.

Unfortunately that PR is light on detail about what the purpose of the resources attribute is, there's no linked issue or any description of what problem it is fixing. 

I also tried adding a lifecycle/ignore_changes to `resources`, but that failed with:
```
│ Adding an attribute name to ignore_changes tells Terraform to ignore future changes to the argument in configuration after the object has been created, retaining the value originally configured.
│
│ The attribute resources is decided by the provider alone and therefore there can be no configured value to compare with. Including this attribute in ignore_changes has no effect. Remove the attribute from ignore_changes to quiet this warning.
```

So I would like to know:
1. What is the intended purpose of adding the resources attribute?
2. If it is indeed only intended as a read-only attribute then does it make sense for changes to show up in a plan?
3. Is there any way we can restore our previous functionality without being pinned to provider 3.0.2, aka essentially ignore changes to `resources`, or even to specific sub-paths under `resources`? 

### Terraform, Provider, Kubernetes and Helm Versions

```
Terraform version: v1.15.3
Provider version: 3.1.1
Kubernetes version: v1.35.3 (eks)
```

### Affected Resource(s)

- helm_release

### Steps to Reproduce
1. `terraform plan`

### References

- [GH-1693](https://github.com/hashicorp/terraform-provider-helm/issues/1693)

### Community Note

[...]

PR #1789, opened 2026-05-12
Fix set_sensitive redaction for keys containing escaped dots (#737 && #1788)

atmask:
## Rollback Plan

If a change needs to be reverted, we will publish an updated version of the library.

## Changes to Security Controls

No changes to security controls. This is a bug fix that corrects existing sensitive value redaction to work with escaped dot characters in key names.

### Description
When using set_sensitive with key names containing literal dots (escaped with `\\.` in HCL, e.g., `server.config.oidc\.config`) the cloaking logic fails to redact the value in metadata output. The root cause is that `cloakSetValue()` and `cloakSetValueModel()` use `strings.Split(valuePath, ".")` which splits on all dots, including escaped ones. This means the function can't traverse the map to the correct key, and the sensitive value leaks in plaintext.                                                                           

The fix introduces a `splitKeyPath()` helper that tokenizes the path while respecting `\.` escape sequences, matching the behavior of Helm's `strvals` package which correctly handles escapes when setting values.  

### Acceptance tests
- [x] Have you added an acceptance test for the functionality being added?

### Release Note
Release note for [CHANGELOG](https://github.com/hashicorp/terraform-provider-helm/blob/main/CHANGELOG.md):
```release-note
Fix `set_sensitive` redaction for keys containing escaped dots (`\.`)
```
### References
  - #737                                                                                              
  - #746             
  - #1221 
  - #1788

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

PR #1737, opened 2026-01-09
Suppress verbose, insecure metadata output in terraform plans

ndebuhr:
Hi team. Taking a stab at the security/UX issue discussed in https://github.com/hashicorp/terraform-provider-helm/issues/1315.

## Rollback Plan

If a change needs to be reverted, we will publish an updated version of the library.

## Changes to Security Controls

**None**

### Description

Keep existing state metadata in plans instead of marking it as unknown when inputs change. This prevents voluminous and low-value computed metadata from cluttering plan output while maintaining correct state updates during `apply`. 

### Acceptance tests
- [X] Have you added an acceptance test for the functionality being added?

Yes, took a stab at that as well

### Release Note
Release note for [CHANGELOG](https://github.com/hashicorp/terraform-provider-helm/blob/main/CHANGELOG.md):

```release-note
Suppress verbose, and potentially insecure, metadata outputs in terraform plans
```
### References

https://github.com/hashicorp/terraform-provider-helm/issues/1315

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

PR #1877, opened 2026-09-03
resource/helm_release: fix experiments.manifest plan noise, apply crashes, and secret redaction

### LINKED ISSUE #1315: Suppress metadata changes at terraform plan?
Older helm provider versions didn't show this `metadata` changes when terraform detected a change/run plan, and the new helm provider versions are posting a wall of text of `metadata` changes which have no real value (to me) and just clog up my tf plan output.

example doing a tf plan on a helm resource where we only updated the image tag var:

```
  # module.api-eks.helm_release.app will be updated in-place
  ~ resource "helm_release" "app" {
        id                         = "api"
      ~ metadata                   = [
          - {
              - app_version = "latest"
              - chart       = "api"
              - name        = "api"
              - namespace   = "production"
              - revision    = 1420
              - values      = jsonencode(
                    {
                      - appConfig          = {
                          - apiBackendReplicationTaskId       = "none"
                          - applicationMode                   = "none"
                          - baseApiUrl                        = "none"
                          - something = "else"
                          - foo = "bar"
                          - 
                          < it goes on for many >
                          < many >
                          < lines >
                          < and it's of no value >
                          < just noise on tf plan >
                          
```

let's say we only have a field  `deploymentTimestamp` updated.  We'd rather see the changed field only on terraform plan, and suppress the whole `metadata` update, e.g. terraform plan should only show

```
  # module.api-eks.helm_release.app will be updated in-place
  ~ resource "helm_release" "app" {
        id                         = "api"
        [...]
        # (25 unchanged attributes hidden)

      + set {
          + name  = "deploymentTimestamp"
          + value = "19012024-225905"
        }

        # (62 unchanged blocks hidden)
        
```
This way the terraform plan is clear and concise, more human (easier to read/follow) without the metadata removal. Does it make sense?

## Terraform version, Kubernetes provider version and Kubernetes version

```
Terraform version: v1.6.5
Helm Provider version: v2.12.0 (same on v2.12.1)
Kubernetes version: v2.24.0
```
## Terraform configuration

```hcl
resource "helm_release" "app" {
  namespace       = var.namespace != "" ? var.namespace : terraform.workspace
  chart           = var.chart_name
  version         = var.chart_version
  name            = var.app_name
  timeout         = var.deployment_timeout
  cleanup_on_fail = var.helm_cleanup_on_fail
  atomic          = var.helm_atomic_creation
  max_history     = var.helm_max_history
  wait            = var.helm_wait_for_completion

  dynamic "set" {
    for_each = local.k8s_app

    content {
      name  = set.key
      value =  set.value
    }
  }

  values = var.some_ingress_values
}

```

## Question

Is there any way to suppress the `metadata` changes at terraform plan?

johannesgesenhues:
## Summary

Six commits from an audit of the `metadata` plan/apply path (always on) and the opt-in `experiments.manifest` / `experiments.keyed_lists` experiments: two plan-noise fixes, three correctness bugs (one a hard apply crash, two parts of one silent secret leak), one new opt-in diff-shape improvement, and a small key-accuracy correction. Item 6 is the result of an independent adversarial review of this branch's own diff, done before opening this PR, which found a real gap in item 4's fix and a test-quality issue in item 3's — both closed here rather than left as follow-ups. Each item below states its own root cause, fix, and verification — no shared narrative to re-derive. All `experiments.*`-gated changes only affect users who opt in; the `metadata`-planning changes (rows 1 and 2) are always on but change plan *content*, not schema shape, so a plan with no experiments enabled should be byte-identical before and after this branch.

| # | Attribute | Change | Issue |
|---|---|---|---|
| 1 | `metadata` (always on) | Stop marking the whole `metadata` object unknown on every relevant change; plan real values, leave only `revision`/`last_deployed`/`notes` unknown | Fixes #1315 |
| 1 | `experiments.manifest` | New `experiments.keyed_lists`: store k8s list-map fields (`containers`, `env`, `ports`, `volumes`, …) keyed by identity instead of array index | Related to #840 |
| 2 | `metadata` (always on) | Fix a forced-replace (rename / move-namespace) planning the *old*, about-to-be-destroyed release's `first_deployed` onto the new one; a nil-pointer panic in an intermediate fix attempt was caught in the same pass | Part of #1315 |
| 3 | `resources[...]` | Fix a hard "Provider produced inconsistent result after apply" crash: the dry run never replicated Helm's own `managed-by=Helm` label stamp | Related to #1791 (different bug); not #1402 |
| 4 | `manifest` / `resources` | Fix `set_sensitive` values never actually being redacted (silent plaintext leak into state/plan); `resources` had *no* redaction mechanism at all | Not previously tracked — see note vs. #1788/#1376 below |
| 5 | `experiments.keyed_lists` | Fix 3 list-map keys that didn't match Kubernetes' own `+listMapKey` declarations; extend keying to `resources` | Groundwork / cleanup |
| 6 | `manifest` / `resources` | Fix a JSON-escaping gap in item 4's redaction (quotes/backslashes/newlines - e.g. a PEM key - weren't redacted); replace a third-party OCI chart test dependency with a local fixture | Closes the limitation disclosed in item 4 |

---

<details>
<summary><b>1. Readable <code>metadata</code> plans + new <code>experiments.keyed_lists</code></b> — Fixes #1315</summary>

**Root cause.** `ModifyPlan` set the entire `metadata` object to unknown whenever `recomputeMetadata()` decided anything relevant had changed. `metadata` carries its own embedded copy of the release's merged values, so a one-line values edit or a chart-version bump reprinted the *entire* previous values blob as removed, with every field marked `(known after apply)`. On a real deployment bumping a chart from `1.83.3` to `1.98.0`, this alone accounted for hundreds of the plan's ~1600 lines for a single actual change — matching the shape of noise reported in #1315 (one commenter there measured 2183/2392 plan lines as this same noise).

**Fix.** Most of `metadata` is knowable before apply: name/namespace come from configuration, chart coordinates and merged values come from the dry run `experiments.manifest` already performs, and `first_deployed` doesn't move on an upgrade. `plannedMetadata` now builds real values for all of that, leaving only `revision`, `last_deployed`, and rendered `notes` unknown — the fields that genuinely can't be known pre-apply. The planned-values string is produced by the same helper the read path uses, so plan and post-apply state can't disagree on it. Write-only (`set_wo`) values keep their field unknown, matching existing read-path behavior.

**New feature (`experiments.keyed_lists`).** Terraform diffs a JSON array by position. A chart adding one `env` entry near the front of a list made every later entry pair against its new neighbour and read as a full-list rewrite:

```
~ { ~ name  = "AWS_DEFAULT_REGION" -> "LITELLM_LOG"
    ~ value = "eu-central-1"       -> "INFO" },
- { name = "BEDROCK_REGION_NAME" ... }   # not actually being removed
```

Kubernetes itself declares these fields `x-kubernetes-list-type: map` — unordered, merged by key. The new opt-in `experiments.keyed_lists` flag re-encodes these fields in `manifest` (and, per item 5, `resources`) as objects keyed by their identifying field instead of arrays, so an insertion reads as one addition. It's opt-in because it changes the shape of the attribute for anyone reading it with `jsondecode()`. Genuinely ordered lists (`args`, `command`) and any list whose entries lack a unique identifying field are left as arrays. Direct precedent: #840 ("improve diff for lists/arrays"), closed by the stale-bot after one automated warning with no maintainer rejection on record.

**Verified.** `TestRecomputeMetadata`, `TestPlannedMetadata_*` (write-only values, nil-state cases), plus ~40 new unit cases across resource kinds, JSON value kinds, and malformed/empty/multi-doc manifests (`manifest_json_kinds_test.go`, `manifest_json_values_test.go`, `manifest_json_malformed_test.go`). Acceptance: `TestAccResourceRelease_recomputeMetadata` against a live cluster.

**Heads-up for review:** #1737 (open, 54 👍, maintainer-engaged) targets the same underlying complaint via a different, opt-in mechanism (flags that blank `metadata.notes`/`metadata.values` to `""`), touching the same `setReleaseAttributes` metadata block and `provider.go` experiments schema as this commit. Not mutually exclusive in intent, but landing both around the same time will likely conflict in those two files — worth flagging now rather than at merge time. #1827 (open, unreviewed) is narrower, fixing only the no-op-replan case; this commit supersedes it in scope rather than conflicting with it.
</details>

<details>
<summary><b>2. Fix a <code>metadata</code> replace-identity bug on forced replacement</b> — part of #1315</summary>

Found while deliberately stress-testing item 1, not by report.

**Root cause.** `plannedMetadata` carried `first_deployed` from `ModifyPlan`'s prior-state argument whenever it was non-nil. But `name` and `namespace` both carry `RequiresReplace()` — on a forced replacement, that prior state describes the release *about to be destroyed*, not the one being planned. Renaming a release or moving its namespace could plan the old, doomed release's `first_deployed` timestamp onto the new one.

**Fix.** An explicit same-identity check, plus a state-nil guard. Verifying that guard mattered surfaced something worse than the wrong-value bug being hunted: a first fix attempt that dropped the same-identity check along with the nil guard crashed the provider with a nil pointer dereference. The shipped fix carries both checks.

**Verified.** `TestAccResourceRelease_replaceDoesNotInheritMetadata` forces a replacement via rename against a live cluster and asserts, at plan time via `plancheck.ExpectUnknownValue`, that `metadata.first_deployed` is unknown — a post-apply state check can't see this bug, since the value is legitimately known again by then. Both the unit test and the acceptance test independently caught the nil-pointer regression when it was deliberately reintroduced to confirm the tests weren't vacuous.

Related to #1315: a later comment on that issue (deadlysyn) reports "Provider produced inconsistent result after apply" on `metadata.last_deployed`/`metadata.revision` when using `lifecycle { ignore_changes = [metadata] }` under provider v3.x. That's a different trigger (an `ignore_changes` interaction, not a bare forced replace) than the bug fixed here, so this is noted as suggestive rather than claimed as fixed.
</details>

<details>
<summary><b>3. Fix <code>resources[...]</code> "inconsistent result after apply" crash on real charts</b> — related to #1791, not #1402</summary>

Found by testing `experiments.manifest` against a real chart (`oci://ghcr.io/berriai/litellm-helm`) across a real version upgrade, not a constructed fixture. Reproduces identically on an unmodified upstream `v3.3.0` checkout — pre-existing, not introduced by items 1–2:

```
Error: Provider produced inconsistent result after apply
.resources["configmap/v1/.../litellm-config"]: was
cty.StringVal("{...\"metadata\":{\"name\":\"litellm-config\",...}}"),
but now
cty.StringVal("{...\"metadata\":{\"labels\":{\"app.kubernetes.io/managed-by\":
\"Helm\"},\"name\":\"litellm-config\",...}}")
```

**Root cause.** Every real Helm Install/Upgrade/Rollback runs Helm's own `setMetadataVisitor` (`helm.sh/helm/v3/pkg/action/validate.go`, `force=true`), which unconditionally stamps `app.kubernetes.io/managed-by=Helm` onto every resource — including ones whose chart template never sets it, which is most of them. `getDryRunResources`'s server-side-apply merge never replicated that stamp, so the planned value for any such resource permanently disagreed with the live read-back post-apply: not a diff, a hard crash on every Create/Update of it.

**Fix.** `setDryRunOwnershipMetadata` applies the same stamp (via the same `k8s.io/apimachinery/pkg/api/meta` accessor `validate.go` uses) to the dry-run object before the merge, local to that one call — nothing here touches the real Install/Upgrade path.

**Verified.** Against a live kind cluster with the real chart across the same version jump: reproduces on a separate, unmodified v3.3.0 worktree, resolved after the fix. `TestSetDryRunOwnershipMetadata` covers the stamping logic with a synthetic fixture. `TestAccResourceRelease_manifestServerDiff` (exercises genuine drift detection) still passes, confirming the fix doesn't mask real differences beyond the one Helm itself always injects.

**Not #1791.** That issue's primary report (an `image` tag continuously patched by an external CI pipeline, drifting against the chart's own value) traces to `getDryRunResources` predicting via server-side-apply with forced conflict resolution — which unconditionally reclaims any field the chart *declares* — while Helm's real apply computes a three-way strategic-merge patch that only touches fields the chart itself changed between versions. Different algorithms, different bug; this commit doesn't touch `ForceConflicts`/`ServerSideApply` and doesn't fix or regress it. (Two follow-up reports on #1791 — a Vault-injected `caBundle`, an AKS webhook's injected `namespaceSelector` — are a third mechanism again: async third-party controllers writing post-apply, outside any Helm apply path.) An unrelated open PR, #1863, adds an `enable_resources` opt-out that sidesteps this whole class of problem rather than fixing the merge-algorithm mismatch; no overlap with this commit.

**Not #1402** (closed, stale-bot, unresolved): that report predates the `resources` attribute entirely (introduced by #1693, merged after #1402 was auto-closed) and its crash is on the flat `.manifest` string via a structurally different code path with no server-side-apply merge on either the plan or read side.
</details>

<details>
<summary><b>4. Fix <code>set_sensitive</code> values never redacted in <code>manifest</code>/<code>resources</code></b> — security-relevant, see limitation below</summary>

Found via adversarial audit of this code path, then confirmed present identically on an unmodified v3.3.0 checkout — pre-existing.

**Root cause.** `redactSensitiveValues(text string, sensitiveValues map[string]string)` iterated the map's *keys* as the strings to search for and mask. All three real call sites built that map backwards — keyed by the `set_sensitive` attribute **name**, with the actual secret value discarded or replaced by a placeholder:

```go
sensitiveValues[set.Name.ValueString()] = "(sensitive value)"   // read path
valuesMap[set.Name.ValueString()] = set.Value.ValueString()      // both plan-time dry runs
```

So the function searched the rendered manifest text for the literal string `"dbPassword"` and never touched the real secret, which flowed into Terraform state and plan output in the clear. This is silent, not a crash: reproducing the original shape and running a real `apply` succeeds with no error, because the (wrong) planned value and the (wrong) post-apply state agree with each other — Terraform's own inconsistent-result check never fires. Only inspecting the raw state file surfaces it.

**Second, independent bug found while writing the regression test.** `getDryRunResources`/`getLiveResources` had no `set_sensitive` redaction at all — the only redaction there, `redactSecretData`, only hashes `Secret.Data`, not a sensitive value appearing in any other resource's fields (an annotation, an env var, a ConfigMap entry).

**Fix.** `redactSensitiveValues` now takes `[]string` — the real values — making the old name-keyed orientation impossible to reintroduce by construction. A new `sensitiveSetValues` helper replaces the duplicated inline logic at all three call sites, filtering null/unknown/empty values (an empty string would make `strings.ReplaceAll(text, "", marker)` match every position and corrupt the output). `sensitiveValues` is threaded through `mapRuntimeObjects`, the one point both the live and dry-run paths funnel through, so `resources` is now covered from the same call.

**Verified.** `TestAccResourceRelease_manifestRedactsSetSensitive` inspects the real on-disk Terraform state after a real `apply`, scoped to the `manifest`/`resources[...]` attributes specifically (not the state blob as a whole — `set_sensitive`'s own value is legitimately present elsewhere in state as a normal input argument; `Sensitive: true` only masks CLI/plan console output). Confirmed non-tautological by deliberately reintroducing the original bug shape at all three call sites: the test correctly catches the plaintext secret; restoring the fix, it passes clean. `TestSensitiveSetValues` and `TestRedactSensitiveValues_EndToEnd` cover the helper directly. `TestAccResourceRelease_manifestServerDiff` still passes.

**A second, narrower gap in this same fix — found by an independent adversarial review during PR prep, before submission — is fixed by item 6 below**: the redaction was a literal `strings.ReplaceAll` of the raw configured value against the already-JSON-marshaled text, so a secret containing a character JSON escapes (a quote, a backslash, or a newline — i.e. most multi-line secrets, such as a PEM private key) didn't byte-match its escaped form and went unredacted too. Not left as a disclosed limitation — closed in the same PR, see item 6.

**Not the same bug as #1788 / #1376 / #1464** (open), despite the surface similarity of "a `set_sensitive` value ends up unredacted." Those report a narrower, orthogonal bug in `cloakSetValue()`/`cloakSetValueModel()` — a different function, used only for the always-on `metadata.values` field — which walks the parsed values map by `strings.Split(valuePath, ".")` to find a leaf to mask by dot-path, and doesn't respect Helm's own `\.`-escape convention, so a name like `config.oidc\.clientSecret` splits into the wrong segments and silently no-ops. That function's value/name orientation is correct for its job; its bug is purely escape-aware tokenizing. An unmerged fix already exists at #1789 (adds an escape-aware `splitKeyPath()`). This PR neither fixes nor regresses it.
</details>

<details>
<summary><b>5. Correct <code>keyed_lists</code> key allowlist, extend to <code>resources</code></b></summary>

**Key-allowlist corrections**, verified against `k8s.io/api` source and its `+listMapKey` declarations:
- `sysctls` removed from the allowlist entirely — `PodSecurityContext.Sysctls` is declared `+listType=atomic` (Kubernetes replaces the whole list on any change, it does not merge by element), so keying it as a map misrepresented what apply actually does, even though every element happens to carry a unique `name`.
- `volumeMounts`/`volumeDevices`: key changed from `name` to the real `mountPath`/`devicePath`.
- `ports` deliberately keeps `name` as a documented heuristic rather than Kubernetes' actual compound key (`containerPort`/`port` + `protocol`), since a single-field key can't express a compound one; duplicate names fall back to an array rather than colliding.

**Extended to `resources`.** `manifest` and `resources` render the same per-resource JSON and suffer identical positional-array diff noise, but `keyed_lists` previously only reached `manifest` — `resources` has its own independent encoding path that nothing had wired the experiment into. Threaded `keyLists` through `mapRuntimeObjects` → `mapResources` → `getLiveResources`/`getDryRunResources`.

**Minor consolidation found along the way:** a duplicate unmarshal/normalize pass in `getDryRunResources`/`getLiveResources` removed (already performed by `mapRuntimeObjects`; `normalizeK8sObject` is idempotent); `setDryRunOwnershipMetadata` simplified to stamp only the label (the two `meta.helm.sh/*` annotations it also stamped were dead code — `stripHelmMetaAnnotations` strips them from both sides before comparison); `convertYAMLManifestToJSON`'s Secret handling now calls the existing `redactSecretData` instead of duplicating its hashing loop; `recomputeMetadata`'s checklist gained the write-only value revision (a `set_wo` bump changes what's applied even though nothing else on the checklist covers it); the Create path's dry-run success branch now calls `plannedMetadata` for parity with Update.

**Verified.** New key-collision tests (two `volumeMounts` of the same `name` at different `mountPath`s — would have collided under the old key), plus `volumeDevices`/`devicePath` and a `ports` heuristic test documenting the fallback-to-array case honestly. `TestAccResourceRelease_keyedListsAppliesToResourcesToo` proves `resources[...]` keying end to end against a live cluster.
</details>

<details>
<summary><b>6. Fix a JSON-escaping gap in item 4's redaction; drop item 3's test's OCI dependency</b> — found during pre-submission review</summary>

Found by an independent, adversarial review agent given the diff with no prior context of this branch's work, run as a final check before opening this PR — the highest-value kind of finding this whole process was built to surface, and reproduced directly before trusting it.

**The redaction gap.** `redactSensitiveValues` compared the raw configured secret value against the already-`json.Marshal`ed manifest/resources text. Any value containing a character JSON escapes — a double quote, a backslash, or a control character such as a newline — never appears in that text as its raw bytes, so the comparison matched nothing and the secret (readable, merely re-escaped) shipped unredacted. Multi-line secrets — a PEM private key or certificate being the obvious real case — were affected on every apply, independent of the map-orientation bug item 4 fixes.

**Fix.** `jsonEscapedForm` returns exactly what a value looks like once embedded in a JSON string field — `json.Marshal`'s own encoding, quotes stripped — and `redactSensitiveValues` now searches for that instead of the raw value. For a plain alphanumeric secret (everything the existing suite already covered) the escaped form is byte-identical to the raw one, so this is purely additive; confirmed every JSON-producing call site in this codebase uses plain `json.Marshal` (not a custom HTML-safe-escaping-disabled encoder), so the escaping rules this lookup assumes are exactly the rules the producers actually use.

**The OCI-chart test dependency.** The same review flagged `TestAccResourceRelease_manifestOCIChartUpgrade`'s dependency on a third-party registry. Investigating rather than just swapping it out: this repo's own `test-chart` genuinely *cannot* exercise item 3's fix — its `_helpers.tpl` already sets `app.kubernetes.io/managed-by` itself, so dry-run and live agree for a reason unrelated to the bug. A new two-file fixture chart, `testdata/charts/bare-metadata`, whose ConfigMap deliberately declares no labels of its own, closes that gap without needing a real chart at all. The OCI-based test is kept alongside it (still `testing.Short()`-gated) for the real-world-fidelity coverage a synthetic chart can't provide — this is additional coverage, not a replacement.

**Verified.** Reproduced the escaping gap directly (a value with embedded newlines/quotes/backslashes survived "redaction" fully readable) before fixing it. Confirmed non-tautological for both fixes: reintroducing the raw-value search fails all 7 new special-character test cases; reintroducing the missing label-stamp reproduces the exact original "Provider produced inconsistent result after apply" crash against the new local fixture chart. Both restored and re-verified passing. New tests: `TestJSONEscapedForm`, `TestRedactSensitiveValues_MultiLineAndSpecialCharacters`, `TestAccResourceRelease_manifestRedactsQuoteContainingSetSensitive` (live-state check, mirroring item 4's), `TestAccResourceRelease_ownershipMetadataLocalChart`.

One thing this fix does *not* attempt: `set_sensitive`'s default (`auto`) and `string` types pass the value through Helm's own `strvals.ParseInto` — the same mini-language `helm --set` uses, where a bare backslash is a structural escape character — so a value containing a literal backslash or newline can already be altered by Helm itself before it ever reaches redaction, independent of anything in this PR. That's pre-existing `set_sensitive` behavior, not touched here; `type = "literal"` bypasses it. Noted for completeness, not filed as part of this PR.
</details>

---

## Security note

Item 4 (redaction) is the one change with real security stakes — it closes a silent secret-leak path where `set_sensitive` values were written to Terraform state and plan output in the clear. Both bugs it fixes (the map-orientation bug, and the resources-attribute gap) only affect users who opt into `experiments.manifest` (default `false`). A second, narrower gap in the same redaction — values containing JSON-special characters, e.g. a multi-line PEM key — was found during pre-submission review and is closed by item 6 in the same PR, not left as a follow-up.

## Related issues not fixed here

- **#1788 / #1376 / #1464** (open) — `cloakSetValue` escaped-dot bug on `metadata.values`. Different function and attribute from item 4's fix; an unmerged fix already exists at #1789. Not touched by this branch.
- **#1791** (open) — `resources[...]` image-tag drift from server-side-apply + `ForceConflicts` vs. Helm's real three-way-merge-patch apply. Different mechanism from item 3's missing-label fix; not addressed here. #1863 (open) adds an `enable_resources` opt-out for this issue and isn't affected by anything in this PR.
- **#1402** (closed, stale-bot, unresolved) — predates the `resources` attribute; out of scope structurally, not just by timing.
- **#1737** (open, 54 👍, maintainer-engaged) — a different, opt-in mechanism touching the same files as item 1. Not mutually exclusive, but flagging the merge-conflict risk for whichever lands second.
- **#1827** (open, unreviewed) — narrower predecessor of the same `metadata`-unknown problem (only the no-op-replan case). Superseded in scope by item 1, not a conflict.

## Test plan

```sh
go build ./...
go vet ./...
gofmt -l .
go test -race ./...
```

Acceptance (needs `TF_ACC=1` and a reachable cluster, e.g. `kind`):

```sh
TF_ACC=1 go test ./helm/... -v -timeout 45m -run \
  'TestAccResourceRelease_(recomputeMetadata|replaceDoesNotInheritMetadata|manifestServerDiff|manifestRedactsSetSensitive|keyedListsAppliesToResourcesToo|manifestOCIChartUpgrade|manifestRedactsQuoteContainingSetSensitive|ownershipMetadataLocalChart)'
```

- For the two most severe fixes (item 2's replace-identity bug, item 4's redaction bug), verification wasn't just "add a test and watch it pass": the original bug shape was deliberately reintroduced at each affected call site to confirm the new regression tests actually fail against it, then reverted and re-verified passing — noted per-item above.
- `TestAccResourceRelease_manifestOCIChartUpgrade` pulls a real chart from a third-party OCI registry (`oci://ghcr.io/berriai/litellm-helm`) across the specific version pair that originally surfaced item 3's bug; it's gated behind `testing.Short()` so it's skipped by default, and kept for the real-world-fidelity coverage a synthetic chart can't provide. Per item 6, a fully local, offline-safe equivalent (`TestAccResourceRelease_ownershipMetadataLocalChart`, using a new two-file fixture chart, `testdata/charts/bare-metadata`) now also covers this bug without any external dependency — this repo's own `test-chart` can't exercise it at all, since its `_helpers.tpl` already sets `app.kubernetes.io/managed-by` itself, so the dry-run and live sides agree for a reason unrelated to the fix.
- Full local run: acceptance tests pass together, including `manifestServerDiff`, confirming none of this masks genuine drift detection. Maintainers should still run the full acceptance suite against a real cluster before merging, per this repo's contribution guidelines.

## Rollback plan

Standard revert; no data migration. `metadata` plan *content* changes (same schema/shape) for all users; `experiments.keyed_lists` is new and opt-in; `experiments.manifest`'s `resources`/`manifest` attributes keep their existing shape except when `keyed_lists` is also enabled (a one-time diff on first enabling that experiment, unchanged from its original behavior).

## Release notes

```release-note:bug
resource/helm_release: Stop replacing the whole `metadata` object with an unknown value whenever the chart version or values change, burying small changes in hundreds of lines of plan output.
```
```release-note:bug
resource/helm_release: Fix `metadata.first_deployed` being planned from the outgoing release's state on a forced replacement (name/namespace change).
```
```release-note:bug
resource/helm_release: Fix `experiments.manifest` planning `resources[...]` values that permanently disagreed with the live cluster for resources whose chart template doesn't itself set Helm's ownership label, causing "Provider produced inconsistent result after apply".
```
```release-note:bug
resource/helm_release: Fix `set_sensitive` values never actually being redacted from the `manifest` and `resources` attributes when `experiments.manifest` is enabled - including a second, narrower gap where a value containing a JSON-special character (a quote, backslash, or newline - as in a multi-line PEM key) wasn't matched against its JSON-escaped form and went unredacted too.
```
```release-note:enhancement
provider: Add the `keyed_lists` experiment, storing Kubernetes list-map fields in `manifest`/`resources` keyed by identity instead of array position.
```
```release-note:bug
provider: Correct three `keyed_lists` list-map keys against Kubernetes' own `+listMapKey` declarations (`sysctls`, `volumeMounts`, `volumeDevices`).
```

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dduzgun-security:
This should close https://github.com/hashicorp/terraform-provider-helm/issues/1376 
cc: @james-warren0

